PR #8, opened 2024-06-30
Add a List example

huacnlee:
https://github.com/huacnlee/gpui-component/assets/5518/2554bb6c-94f6-4808-87d6-9a04d7421b26

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1147,6 +1147,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


reyamir:
Oh, thank you, I'm looking an example for list with gpui 😅 too.
Do you know how to do list with async data? I'm still not figuring render an item by async query in gpui yet.

huacnlee:
@reyamir Check my last commit.

PR #13, opened 2024-07-03
Add SvgImg to render svg, and with a Chart example

huacnlee:
https://github.com/huacnlee/gpui-component/assets/5518/576da9f9-4bff-4faf-a17e-19860cf6f948


huacnlee:
image

PR #93, opened 2024-07-31
StoryContainer: rename add_pane to add_item

stormasm:
It seems to me that the current method name of *add_pane* is confusing as you
are not actually adding a new pane but rather we are

- adding an item to the pane
- and/or updating the pane

```rust
       cx.spawn(|mut cx| async move {
            pane.update(&mut cx, |pane, cx| {
                let view = cx.new_view(|cx| Self::new(name, description, cx).story(story));

                pane.add_item(Box::new(view.clone()), true, true, None, cx);
                view
            })
        })
 ```

### DIFF
--- a/crates/app/src/story_workspace.rs
+++ b/crates/app/src/story_workspace.rs
@@ -41,7 +41,7 @@ impl StoryWorkspace {
         })
         .detach();
 
-        StoryContainer::add_pane(
+        StoryContainer::add_item(
             "Buttons",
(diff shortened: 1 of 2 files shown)


stormasm:
So the other possible name instead of *add_item* would be *update_pane*

stormasm:
another option could be 

*add_item_to_pane*

huacnlee:
We not need to change this, that currently just a demo, I may change it again and again, so just don't mind this name.

PR #169, opened 2024-08-19
Set cursor width to an integer to prevent width fluctuation on low PPI devices.

eastack:
Before

https://github.com/user-attachments/assets/38ceac8e-c98d-43c7-844b-0a2fb7d416a5

After

https://github.com/user-attachments/assets/e3f9c70e-773e-4606-ad6a-021b0e7fd7c9


huacnlee:
Thank you.

There actually have a lot of same issue in GPUI. We need to keep all px in integer to avoid that.

PR #183, opened 2024-08-28
table: Add `load_more` delegate method and called when scrolled near the bottom.

huacnlee:
https://github.com/user-attachments/assets/03001a21-6a3d-4c06-a00c-9dea8208e357


huacnlee:
Closes #182

PR #221, opened 2024-09-05
Add ButtonGroup

xda2023:
image

### DIFF
--- a/crates/ui/src/button.rs
+++ b/crates/ui/src/button.rs
@@ -11,6 +11,19 @@ use gpui::{
     RenderOnce, SharedString, StatefulInteractiveElement as _, Styled, WindowContext,
 };
 
+pub enum ButtonBorderSide {
+    All,
+    NoLeft,
+    NoRight,
+    NoLeftRight,
+}
+
+pub enum ButtonRoundedSide {
+    All,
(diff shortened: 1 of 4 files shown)


crates/ui/src/button.rs
```diff
@@ -109,6 +122,8 @@ pub struct Button {
     selected: bool,
     style: ButtonStyle,
     rounded: ButtonRounded,
+    rounded_side: ButtonRoundedSide,
```
huacnlee: Use `gpui::Corners`

huacnlee: https://github.com/zed-industries/zed/blob/c20c2cf380f126487a58c5698df0a3ab98840d50/crates/gpui/src/geometry.rs#L1854

crates/ui/src/button.rs
```diff
@@ -109,6 +122,8 @@ pub struct Button {
...
+    border_side: ButtonBorderSide,
```
huacnlee: use `gpui::Edges`

https://github.com/zed-industries/zed/blob/c20c2cf380f126487a58c5698df0a3ab98840d50/crates/gpui/src/geometry.rs#L1484

PR #247, opened 2024-09-16
input: Update Input selection behavior to like OS input.

huacnlee:
Revert old implementation

- https://github.com/huacnlee/gpui-component/pull/26
- https://github.com/zed-industries/zed/pull/14340

## Before

https://github.com/user-attachments/assets/52ce2e3a-3e1d-4ba2-af42-76b1df987404

## After

https://github.com/user-attachments/assets/9aa8cfe1-5928-4d20-b534-806140c7ee59

## The Input in the Browser

https://github.com/user-attachments/assets/c913248e-b1dc-46d6-a517-e1231e606171

### DIFF
--- a/crates/ui/src/input/input.rs
+++ b/crates/ui/src/input/input.rs
@@ -410,12 +410,6 @@ impl TextInput {
         self.is_selecting = false;
     }
 
-    fn on_mouse_move(&mut self, event: &MouseMoveEvent, cx: &mut ViewContext<Self>) {
-        if self.is_selecting {
-            self.select_to(self.index_for_mouse_position(event.position), cx);
-        }
-    }
-
     fn show_character_palette(&mut self, _: &ShowCharacterPalette, cx: &mut ViewContext<Self>) {
         cx.show_character_palette();
     }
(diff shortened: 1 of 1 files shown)


huacnlee:
https://github.com/user-attachments/assets/42fffadd-41a5-419b-bd78-1f65eeba6c02

@ConradIrwin To check out these details, this changes is unlike your was done https://github.com/zed-industries/zed/pull/14350. When this mouse is on the cursor left, then select to left, otherwise select to right. 

And this also listen the mouse event on the window not on Input to continue handle event mouse is moved out the Input bounds.

huacnlee:
Diff Zed and Chrome

https://github.com/user-attachments/assets/d20a225b-705e-411e-9433-1714a13ef484

huacnlee:
And why we need to do like this: 

When we fast move mouse to left, we may end mouse at the Input top side or bottom size, they may both happen. Because we can not move in the horizontal line.

So we need to continue the user move direction to select text. If we move to left, that means we want to select to left, otherwise we want to select to right.

PR #256, opened 2024-09-20
table: Render table stripe on background even there is no enough rows

xda2023:
image

- And add `flex_shrink_0` to table head.

Fix https://github.com/huacnlee/gpui-component/issues/160

### DIFF
--- a/crates/ui/src/table.rs
+++ b/crates/ui/src/table.rs
@@ -742,6 +742,28 @@ where
         let cols_count: usize = self.delegate.cols_count();
         let rows_count = self.delegate.rows_count();
 
+        let row_height = self.vertical_scroll_handle.0.borrow().last_item_height;
+        let total_height = self
+            .vertical_scroll_handle
+            .0
+            .borrow()
+            .base_handle
+            .bounds()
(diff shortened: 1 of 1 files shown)


crates/ui/src/table.rs
```diff
@@ -790,66 +795,117 @@ where
                 } else {
                     this.child(
                         h_flex().id("table-body").flex_grow().size_full().child(
-                            uniform_list(view, "table-uniform-list", rows_count, {
+                            uniform_list(view, "table-uniform-list", rows_count + 100, {
```
huacnlee: Why plus 100?

xda2023: I think the best implementation is to change the internal of uniform_list, but uniform_list is in the zed codebase, so I didn't go this way. This implementation now assumes that uniform_list cannot be changed, so +100 lines of fake rows are added and scrolling to the fake part is disabled when scrolling, which has certain limitations.

xda2023: Perhaps let me try changing uniform_list and come up with a different solution

PR #286, opened 2024-09-27
button: Fix Button text to align vertical center.

huacnlee:
image

The point is add `.font_family(".SystemUIFont")` to root view.


huacnlee:
Close https://github.com/zed-industries/zed/issues/17777

PR #294, opened 2024-10-02
scrollbar: Smaller the scrollbar size.

huacnlee:
https://github.com/user-attachments/assets/cf50296b-fc28-4c09-b0fa-5227bb97ce81


huacnlee:
Ref #260

Issue #302, opened 2024-10-03
Saving the Layout is Broken on the latest main


stormasm:
Jason,

You probably already know this :)

But saving the layout is currently not working with your latest fixes in the past day.

A simple test is to reduce the list on the left side of the screen.
Then close out of everything and bring it back up.
And the list appears like it looked before you shrink it down

Everything is working fine at this point...
https://github.com/huacnlee/gpui-component/tree/7de1c6981e22317f00fea629ae32c44b5ef317c6

Thanks ! 
You are doing amazing work on helping us to better understand gpui.
Michael

PS. Once you get it working I will be happy to test again..

huacnlee:
Yes, I will to this. I just merge a part of work.

stormasm:
Thank you very much !
It is now all working beautifully 😄

PR #303, opened 2024-10-03
dock: Add to support save left, right, bottom dock state.

huacnlee:
The dock state have been updated to use `DockAreaState`.


huacnlee:
Fix #302

PR #327, opened 2024-10-10
dock: Fix zoom support for side dock panel.

huacnlee:
- Fix emit `PanelEvent::LayoutChanged` when panel resized.


huacnlee:
Closes #313

PR #341, opened 2024-10-14
drawer: support ESC to close it by on_action and call on_close

### LINKED ISSUE #263: Drawer: support ESC to close it

xda2023:
Fix #263

### DIFF
--- a/crates/ui/src/root.rs
+++ b/crates/ui/src/root.rs
@@ -57,7 +57,14 @@ impl<'a> ContextModal for WindowContext<'a> {
             if root.active_drawer.is_none() {
                 root.previous_focus_handle = cx.focused();
             }
-            root.active_drawer = Some(Rc::new(build));
+
+            let focus_handle = cx.focus_handle();
+            focus_handle.focus(cx);
+
+            root.active_drawer = Some(ActiveDrawer {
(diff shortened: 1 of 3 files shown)


crates/ui/src/root.rs
```diff
@@ -57,7 +57,14 @@ impl<'a> ContextModal for WindowContext<'a> {
             if root.active_drawer.is_none() {
                 root.previous_focus_handle = cx.focused();
             }
-            root.active_drawer = Some(Rc::new(build));
+
+            let focus_handle = cx.focus_handle();
+            focus_handle.focus(cx);
```
huacnlee, line 62: We could't do `focus` here, this will make some unexpected bug.

huacnlee: I was found this in Modal before, it's was also used `focus`, then I found some issue, so it now removed.

Ref: #250

huacnlee: The notification layout not display now.

xda2023: @huacnlee notify layer: The problem with the display not working can be solved by rebasing the changes onto the latest `main` branch. I suspect it has something to do with an update to a dependency version, possibly taffy

PR #365, opened 2024-10-18
Add the Preset Ranges and add border for today

xda2023:
Preset Ranges have been added. The “Today” border has also been added.


huacnlee:
Closes https://github.com/longbridgeapp/gpui-component/issues/166

PR #410, opened 2024-11-12
svg_img: Fix the conversion error of the SVG fill color with alpha

madcodelife:
### Before
image

### After
image


huacnlee:
Ref GPUI fixes https://github.com/zed-industries/zed/pull/20537

PR #424, opened 2024-11-15
input: add InputEvent::Cleaned

stayhydated:


crates/ui/src/input/input.rs
```diff
@@ -399,6 +400,7 @@ impl TextInput {
 
     fn clean(&mut self, _: &ClickEvent, cx: &mut ViewContext<Self>) {
         self.replace_text("", cx);
+        cx.emit(InputEvent::Cleaned);
```
huacnlee, line 403: The `replace_text` is already emitted `InputEvent::Change`, if the argument text is empty that means is clean. 

https://github.com/longbridgeapp/gpui-component/blob/3db84c12472bdd7ff513609ce391539cfc89a633/crates/ui/src/input/input.rs#L790

So I think we not need add another one.

stayhydated: Oh yeah, this makes sense

PR #493, opened 2024-12-16
fix: wrong size variant in sizable trait

reyamir:
### Changelog:
- Adjust `/// Set to Size::Medium` to `/// Set to Size::Large`
- Add size `.medium()`

### DIFF
--- a/crates/ui/src/styled.rs
+++ b/crates/ui/src/styled.rs
@@ -215,17 +215,22 @@ pub trait Sizable: Sized {
     /// Or a `Pixels` to set a custom size: `px(30.)`
     fn with_size(self, size: impl Into<Size>) -> Self;
 
-    /// Set to Size::Small
-    fn small(self) -> Self {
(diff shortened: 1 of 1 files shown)


crates/ui/src/styled.rs
```diff
@@ -215,17 +215,22 @@ pub trait Sizable: Sized {
...
     /// Set to Size::Small
     fn small(self) -> Self {
         self.with_size(Size::Small)
     }
 
-    /// Set to Size::XSmall
-    fn xsmall(self) -> Self {
-        self.with_size(Size::XSmall)
+    /// Set to Size::Medium
+    fn medium(self) -> Self {
```
huacnlee: Remove this, the default is Medium

huacnlee (changes requested):

crates/ui/src/styled.rs
reyamir: I've updated it

PR #495, opened 2024-12-17
virtual_list: Add VirtualList

huacnlee:
Vistual List for render a large number of differently sized rows/columns.

> NOTE: This must ensure each column width or row height.

Unlike the `uniform_list`, the each item can have different size. This is useful for more complex layout, for example, a table with different row height.


reyamir:
I wonder is there a way to apply reverse scroll to virtual list? Like chat app, scroll from bottom to top instead of top to bottom.

huacnlee:
GPUI have a `list` method for this case. Zed have a Chat UI, you can see it.

PR #502, opened 2024-12-19
panel: Add visible method to Panel trait.

huacnlee:
## Added

- Add `visible` method to `Panel` trait, if return false this panel will disappear.

## Break changes

- Renamed `closeable` to `closable`.
- The `title_style`, `closable`, `zoomable` in `Panel` trait have been change `cx: &WindowContext` to `cx: &AppContext`.


huacnlee:
To hide a panel in a dock layout is too complex. Just give up.

PR #524, opened 2025-01-02
dropdown : Modularize creation

FranzJvezNiceAxa:
Makes `new()` simpler and implements the following setters:
- ~~`set_delegate`~~ `delegate`
- ~~`set_selected_index`~~ `index`

old:
```rust
Dropdown::new("dropdown-country", countries, Some(6), cx).cleanable()
```

new:
```rust
Dropdown::new("dropdown-country", cx)
                .delegate(countries, cx)
                .index(Some(6), cx)
                .cleanable()
```

Good use case for this change would be something like the ftl example on `crates/story/src/dropdown_story.rs`

### DIFF
--- a/crates/story/src/dropdown_story.rs
+++ b/crates/story/src/dropdown_story.rs
@@ -156,6 +156,13 @@ impl DropdownStory {
                     Dropdown::new("string-list3", cx)
                         .set_delegate(Vec::<SharedString>::new(), cx)
                         .small()
+                        .empty(|cx| {
+                            h_flex()
+                                .h_24()
+                                .justify_center()
+                                .text_color(cx.theme().muted_foreground)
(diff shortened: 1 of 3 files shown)


huacnlee (changes requested):
The `set_` prefixed method should not return `Self`.

huacnlee:
I think this is not a good change. There is no need to modify the parameters of Dropdown.

Also, looking at the changes in your story, it seems that you made this change to achieve I18n for Dropdown item. This is not the right way. Dropdown's `DropdownDelegate` can be a `trait DropdownItem`, in which the `title` method can achieve this purpose.

We need to do the I18n get text in the `render` stage, so that will keep the localized text to follow the current App locale in runtime. This is also our actually did in our project.

PR #578, opened 2025-01-24
table: Improve sort icon color and fix stripe row display.

huacnlee:
- Improve sort icon color:

    image

- Avoid duplicate border when enabled fixed column.

    image

TODO

- [x] The extra rows not show when `stripe` is true.


huacnlee:
image

PR #601, opened 2025-02-05
Update panel toolbar button type to AnyElement

ihavecoke:


huacnlee (changes requested):
The `toolbar_buttons` must uniform the Button style in `small`.

This change is not good idea.

PR #608, opened 2025-02-07
color_picker Fix popup does not close by clicking color button again.

sectore:
## How to reproduce

1. `cargo run` to run dashboard
2. Click `color-picker` button to open its layer and try to click same button again, to close the layer. The layer won't close. It will close only by clicking outside or selecting a color.

click-cp

## Fix

Move `on_mouse_up_out` handler from inner to upper child to **avoid** a race condition by changing value of `self.open` twice in `on_escape` AND `toggle_picker`.


huacnlee:
Thank you.

PR #686, opened 2025-03-04
scrollbar: Fix Scrollbar not clickable when it is always showing.

### LINKED ISSUE #685: scrollbar: dragging thumb or clicking track do not work before mouse wheel is used once before
## How to reproduce

1. Run any example with a scrollbar, e.g. `cargo run --example list`. 
2. Don't scroll by mouse (wheel), but try to click the `thumb` of the scrollbar to drag it. Or click the `track` of the scrollbar. No scrolling - nothing happened.
3. Now scroll by mouse (wheel) once. Then try to click again (similar to step 2). It will work now.

Image

## First investigation

Check [`scrollbar.rs`](https://github.com/longbridge/gpui-component/blob/8c4884d8818db818234c7ea04b4ec355bda7c21a/crates/ui/src/scroll/scrollbar.rs) -> `is_visible` is still `false` in `paint` ([code](https://github.com/longbridge/gpui-component/blob/8c4884d8818db818234c7ea04b4ec355bda7c21a/crates/ui/src/scroll/scrollbar.rs#L635)), so `MouseDownEvent` ([code from here](https://github.com/longbridge/gpui-component/blob/8c4884d8818db818234c7ea04b4ec355bda7c21a/crates/ui/src/scroll/scrollbar.rs#L635)) can't be handled properly. That's because `dragged_axis` is still `None` in [`is_scrollbar_visible`](https://github.com/longbridge/gpui-component/blob/8c4884d8818db818234c7ea04b4ec355bda7c21a/crates/ui/src/scroll/scrollbar.rs#L166-L179) as long as `ScrollWheelEvent` is **not** fired once.

## Extra info

```sh
rustc --version
rustc 1.85.0 (4d91de4e4 2025-02-17)
```

```sh
OS: Ubuntu 24.04.2 LTS x86_64
```

sectore:
On `platform` level,  [should_auto_hide_scrollbars()](https://github.com/zed-industries/zed/blob/76a81607deec5b9093237cd3e98506dccc9f9565/crates/gpui/src/platform.rs#L205) might be `false` for some `platform`s, e.g. for `Linux`. 

That's why another check is needed to make scrollbars work properly on Linux.

Fixes #685

### DIFF
--- a/crates/ui/src/scroll/scrollbar.rs
+++ b/crates/ui/src/scroll/scrollbar.rs
@@ -4,7 +4,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use crate::ActiveTheme;
+use crate::{ActiveTheme, Theme};
 use gpui::{
     fill, point, px, relative, App, Bounds, ContentMask, CursorStyle, Edges, Element, EntityId,
     Hitbox, Hsla, IntoElement, MouseDownEvent, MouseMoveEvent, MouseUpEvent, PaintQuad, Pixels,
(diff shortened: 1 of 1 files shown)


huacnlee:
Thanks.

PR #744, opened 2025-03-26
Add italian menu

Angelk90:
@huacnlee : When I try to change the language it stays in English, is there something else I need to set?
Screenshot 2025-03-26 alle 16 23 26


huacnlee:
Not need add to example, last PR merge I just considered that useful, but the example not needed included all languages.

Thank you

huacnlee:
Sorry, I am close of this PR for now, later I will think about how to organize more languages or write a guide for this.

PR #758, opened 2025-03-31
input: Add `masked` builder method to Input.

huacnlee:


Angelk90:
@huacnlee : Icon doesn't change though?

huacnlee:
This is mouse down to show password, not the switcher. So it not needs to change icon. There is already have mouse down state.

PR #770, opened 2025-04-02
dock: Fix to sync active state to panels when collapse dock.

JakkuSakura:
This is a continuation of #764
Without this, if we drag webview to sidebar and collapse, webview is not aware of itself being hidden
image

### DIFF
--- a/crates/ui/src/dock/mod.rs
+++ b/crates/ui/src/dock/mod.rs
@@ -410,7 +410,7 @@ impl DockItem {
                 }
             }
             DockItem::Tiles { .. } => {}
-            DockItem::Panel { .. } => {}
+            DockItem::Panel { view } => view.set_active(!collapsed, window, cx),
         }
     }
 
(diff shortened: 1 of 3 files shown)


huacnlee:
You don't have to follow up and insert the WebView into the Panel. This will cause many problems, WebView used into GPUI there have a lot of limit.

Actually in our application, we only use WebView to render into a Drawer or a Modal.

huacnlee:
And I also [removed WebView story](https://github.com/longbridge/gpui-component/pull/769/commits/f63a2a9b6154c76f27b9858b132bbc53da35be61) in PR #769 from the main example application, because if we enable this, the WebView will cause some unexpected problems (For example: It will get the focus handle in high priority, this will break the general element's events).

JakkuSakura:
I see. This PR is clean though. It doesn't add WebView anywhere. It just fixes the relevant issue

huacnlee:
I have done a test, this change is not correct. When I collapse a bottom dock, the log is display like:

```
panel: Image active: true
panel: Icon active: true
panel: Image active: false
panel: Icon active: false
panel: Image active: true
panel: Icon active: true
panel: Tooltip active: true
```

JakkuSakura:
What should be the correct output?

huacnlee:
You can test it, the log is unexpected.

JakkuSakura:
Ah I see. Fixed now

huacnlee:
Thanks.

PR #803, opened 2025-04-17
input: Add `Esc` to clean.

sectore:
By pressing `Esc`:
- unselect text
- clear `input` (optional, set by `clean_on_escape`)

Check updated `input` stories to see examples.

## Demo

input_esc

### Before

before

### Now

after


huacnlee:
I don't think this is correct for every case, this feature should be a feature by user do itself.

sectore:
> I don't think this is correct for every case, this feature should be a feature by user do itself.

It's not for every case. Just if `self.cleanable` is set to `true`. That's if an user see the `clear` button, pressing `ESC` key would be an expected behavior IMO.

huacnlee:
I mean, even based on `cleanable`. You can check same feature on x.com.

sectore:
>You can check same feature on x.com.

Ah I see, `x.com` has a more complex handling as it shows a popup in addition and uses `ESC` to close that popup etc.

> this feature should be a feature by user do itself.

Do you mean something like an extra builder function `cleanable_by_esc(mut self)`? Or by extending `InputEvent::PressEsc` (or `InputEvent::PressKey`) or anything else?

huacnlee:
You can read same behavior in List:

https://github.com/longbridge/gpui-component/blob/b6dd4a3fb431abade62a7397743ffc5174e134c8/crates/ui/src/list/list.rs#L40
https://github.com/longbridge/gpui-component/blob/b6dd4a3fb431abade62a7397743ffc5174e134c8/crates/ui/src/list/list.rs#L407

huacnlee (commented):
I think, or we can have a `reset_on_escape`.

https://github.com/longbridge/gpui-component/blob/b6dd4a3fb431abade62a7397743ffc5174e134c8/crates/ui/src/list/list.rs#L392

crates/story/examples/alert.rs
```diff
@@ -1,9 +1,5 @@
 use gpui::*;
-use gpui_component::{
-    alert::Alert,
-    button::{Button, ButtonGroup},
-    v_flex, IconName, Selectable as _, Sizable as _, Size,
-};
```
sectore, line 6: This change/fix is just because `rust-analyzer` did complain about it.

crates/ui/src/input/input.rs
```diff
@@ -233,6 +235,7 @@ pub struct TextInput {
     pub(super) mask_toggle: bool,
     pub(super) appearance: bool,
     pub(super) cleanable: bool,
+    pub(super) clean_by_escape: bool,
```
huacnlee, line 238: `reset_on_escape`

crates/ui/src/input/input.rs
```diff
@@ -1059,10 +1069,14 @@ impl TextInput {
         cx.notify();
     }
 
-    fn clean(&mut self, _: &ClickEvent, window: &mut Window, cx: &mut Context<Self>) {
+    fn clean(&mut self, window: &mut Window, cx: &mut Context<Self>) {
         self.replace_text("", window, cx);
     }
 
+    fn on_clean(&mut self, _: &Clean, window: &mut Window, cx: &mut Context<Self>) {
```
huacnlee, line 1076: on_action_escape

crates/ui/src/input/input.rs
```diff
@@ -662,6 +666,12 @@ impl TextInput {
         self
     }
 
+    /// Set true to clear the input by pressing Escape key when the input field is not empty.
+    pub fn clean_by_escape(mut self) -> Self {
```
huacnlee: Sorry, I think this is better named: `clean_on_escape`?

crates/ui/src/input/input.rs
```diff
@@ -85,6 +85,7 @@ actions!(
         MoveToPreviousWord,
         MoveToNextWord,
         TextChanged,
+        Clean
```
huacnlee: This action is better rename to `Escape` like `Enter`, because in the future, Esc will have some other functions.

And here you have missed some details, for example we selected something, if press `Escape`, it must cancel selection first.

crates/story/examples/alert.rs
```diff
@@ -1,9 +1,5 @@
...
+
 use story::{AlertStory, Assets};
```
my12345677, line 3: 1234567890

PR #805, opened 2025-04-18
table: Ensure to select row when prev or next.

ylinwind:


crates/ui/src/table.rs
```diff
@@ -589,15 +593,23 @@ where
...
-            if self.delegate.can_loop_select(cx) {
-                selected_row = 0;
-            }
+        let rows_count = self.delegate.rows_count(cx);
+        if rows_count < 1 {
+            return;
         }
 
+        let selected_row = match self.selected_row {
+            Some(selected_row) if selected_row < rows_count - 1 => selected_row + 1,
```
huacnlee: `rows_count - 1` will crash, when `rows_count` is 0.

ylinwind: Processed, changed to `saturating_sub(1)` implementation.

PR #823, opened 2025-04-30
svg_img: No longer use the global image loader.

sunli829:
Update SvgImg load logic and try to drop image to fix VRAM leak issue.


huacnlee:
Close this, by test in project, there still have some issues can't resolve the primary issue.

PR #837, opened 2025-05-07
Add Nix/Flake Development Environment for gpui-component

Losses:
## Overview

This PR adds Nix and Flake support to the gpui-component project, providing developers with a reproducible, declarative development environment that works across various operating systems.

## What is Nix/Flake?

Nix is a powerful package manager and build system that takes a purely functional approach to package management. Unlike traditional package managers, Nix stores packages in isolation from each other in the Nix store, preventing dependency conflicts and ensuring reproducible builds.

Flakes are an extension to Nix that provide:
- Hermetic evaluation: builds are reproducible and depend only on the inputs specified in the flake
- Standardized structure: clear conventions for inputs, outputs, and entry points
- Version pinning: exact dependency versions are locked via the flake.lock file
- Improved composability: flakes can easily depend on other flakes

## Benefits for Developers

By adding Nix/Flake support, this PR provides the following benefits:

1. **Consistent Development Environment**
   - All developers work with identical dependencies and tooling regardless of their OS (Linux, macOS)
   - No more "works on my machine" issues when collaborating

2. **Zero-Configuration Setup**
   - Simply run `nix develop` to enter the development environment with all required dependencies
   - New team members can start contributing immediately without complex setup instructions

3. **Isolated Dependencies**
   - Project dependencies don't conflict with system packages
   - Multiple projects with different dependency versions can coexist peacefully

4. **Reproducible Builds**
   - Build outputs are deterministic and can be reproduced exactly on different machines
   - CI builds match local development builds precisely

5. **Declarative Configuration**
   - All required dependencies and development tools are explicitly declared in flake.nix
   - Environment changes are versioned in Git alongside code

## Implementation Details

The PR adds:
- `flake.nix`: Defines the development environment and its dependencies
- `flake.lock`: Pins exact versions of all dependencies

The environment includes all build dependencies, development tools, and runtime requirements needed for gpui-component development.

## NOTICE

**Important**: This Nix configuration is completely optional and provided as an alternative development setup method. It is primarily intended to benefit:

1. **NixOS Users**: Developers already using NixOS can immediately work with the project without manual dependency installation
2. **Nix Users on Other Platforms**: Developers using Nix on macOS or Linux can leverage the consistent environment

The project remains fully accessible through traditional setup methods. Existing development workflows continue to work as before, and there is no requirement or expectation for team members to adopt Nix.

This addition simply provides an alternative, zero-configuration approach for those who prefer it or are already using Nix. The project's primary build and development instructions remain unchanged.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,53 @@
+{
+  description = "gpui-component";
+
+  inputs = {
+    nixpkgs.url      = "github:NixOS/nixpkgs/nixos-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    flake-utils.url  = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, rust-overlay, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        overlays = [ (import rust-overlay) ];
+        pkgs = import nixpkgs {
+          inherit system overlays;
+        };
+      in
+      {
+        devShells.default = with pkgs; mkShell {
+          buildInputs = [
+            openssl
+            pkg-config
+            eza
+            fd
+            protobuf
+            xorg.libX11
+            glib
+            pango
+            atkmm
+            gdk-pixbuf
+            gtk3
+            libsoup_3
+            webkitgtk_4_1
+            libxkbcommon
+            vulkan-loader
+            (rust-bin.beta.latest.default.override {
+              extensions = [ "rust-src" ];
+            })
+          ];
+
+          shellHook = ''
+            alias ls=eza
+            alias find=fd
+          '';
+
+          env = {
+            RUST_BACKTRACE = "1";
+            LD_LIBRARY_PATH = lib.makeLibraryPath [ vulkan-loader ];
+          };
+        };
+      }
+    );
+}
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,96 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1746461020,
+        "narHash": "sha256-7+pG1I9jvxNlmln4YgnlW4o+w0TZX24k688mibiFDUE=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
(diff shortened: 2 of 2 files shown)


huacnlee:
Do you have passed test, I have get an error:

```
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'nix-shell'
         whose name attribute is located at /nix/store/9rc9abg9f664bjfhzfp4cb8mrwh7b5y4-source/pkgs/stdenv/generic/make-derivation.nix:439:13

       … while evaluating attribute 'buildInputs' of derivation 'nix-shell'
         at /nix/store/9rc9abg9f664bjfhzfp4cb8mrwh7b5y4-source/pkgs/stdenv/generic/make-derivation.nix:494:13:
          493|             depsHostHost = elemAt (elemAt dependencies 1) 0;
          494|             buildInputs = elemAt (elemAt dependencies 1) 1;
             |             ^
          495|             depsTargetTarget = elemAt (elemAt dependencies 2) 0;

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: Package ‘webkitgtk-2.48.1+abi=4.1’ in /nix/store/9rc9abg9f664bjfhzfp4cb8mrwh7b5y4-source/pkgs/development/libraries/webkitgtk/default.nix:254 is marked as broken, refusing to evaluate.

       a) To temporarily allow broken packages, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_BROKEN=1

          Note: When using `nix shell`, `nix build`, `nix develop`, etc with a flake,
                then pass `--impure` in order to allow use of environment variables.

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowBroken = true; }
       in configuration.nix to override this.

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowBroken = true; }
       to ~/.config/nixpkgs/config.nix.
```

Losses:
@huacnlee Yes, this config passed on my side by running `nix develop` and successfully yield a running program, are you compiling it in a Linux distribution or macOS?
image

huacnlee:
macOS

flake.nix
```diff
@@ -0,0 +1,52 @@
...
+          inherit system overlays;
+        };
+      in
+      {
+        devShells.default = with pkgs; mkShell {
+          buildInputs = [
+            openssl
+            pkg-config
+            eza
+            fd
```
huacnlee: I think we can remove `eza`, `fd`. 

And I am not sure is `pango`, `atkmm`, `libsoup_3`, `gdk-pixbuf` is required? 

If that is not needed, we should keep this clean.

Losses:
@huacnlee Thank you for your feedback regarding the Nix configuration on macOS.

I wanted to clarify that configuring Nix on macOS is an extremely complex and relatively niche area. The flake configuration provided in this Pull Request was specifically designed for Linux developers, as that was the primary target platform.

Unfortunately, I don't personally own a Mac device, which makes it difficult for me to debug or test these configurations in a macOS environment. 

Regarding the dependencies you mentioned - pango, atkmm, libsoup_3, and gdk-pixbuf - these are all essential dependencies for compilation on Linux platforms. I've tested removing each one individually, and the compilation fails without them. They aren't optional components but rather core requirements for the build process on Linux.

huacnlee:
Ok, let merge first, I will try to fix them later.

huacnlee:
Thank you

PR #838, opened 2025-05-07
root: Add `AppContextModal` trait for add `open_modal` methods to `gpui::App`.

huacnlee:
This change to support `cx` to open_modal, open_drawer by use `cx.active_window`.

To make sure that we can open modal active current window.

### DIFF
--- a/crates/story/src/modal_story.rs
+++ b/crates/story/src/modal_story.rs
@@ -12,7 +12,7 @@ use gpui_component::{
     h_flex,
     input::TextInput,
     modal::ModalButtonProps,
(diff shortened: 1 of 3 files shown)


huacnlee:
```
thread 'main' panicked at /Users/jason/work/gpui-component/crates/ui/src/root.rs:289:14:
called `Result::unwrap()` on an `Err` value: window not found
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

huacnlee:
Close this, after thinking about it, the method is still not very good.

PR #857, opened 2025-05-15
chore: Unify icon parameter types in ListItem and Dropdown.

ratazzi:
## Changes

This PR unifies icon parameter types in UI components:

- Changed `ListItem.check_icon` parameter type from `IconName` to `impl Into<Icon>`
- Changed `Dropdown.icon` parameter type from `impl Into<IconName>` to `impl Into<Icon>`
- Updated related rendering logic

## Impact

- **API Consistency**: All components now accept the same icon parameter type
- **Backward Compatible**: Existing code using `IconName` continues to work
- **Enhanced Flexibility**: Components can now accept custom `Icon` instances directly

### DIFF
--- a/crates/ui/src/dropdown.rs
+++ b/crates/ui/src/dropdown.rs
@@ -161,7 +161,7 @@ where
 
         if let Some(item) = self.delegate.get(ix) {
             let list_item = ListItem::new(("list-item", ix))
-                .check_icon(Icon::new(IconName::Check))
+                .check_icon(IconName::Check)
                 .selected(selected)
                 .input_text_size(size)
                 .list_size(size)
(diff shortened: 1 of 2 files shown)


crates/ui/src/dropdown.rs
```diff
@@ -161,7 +161,7 @@ where
 
         if let Some(item) = self.delegate.get(ix) {
             let list_item = ListItem::new(("list-item", ix))
-                .check_icon(IconName::Check)
```
huacnlee: Here not need to change, we'd like to keep accepting `IconName` and `Icon` both.

huacnlee: The IconName can into `Icon`.

ratazzi: Thanks @huacnlee! I've changed it back to accept `IconName` as you suggested.

huacnlee: Thanks

Issue #870, opened 2025-05-19
calendar: Support custom matcher impl


0x1997:
for matching local holidays and future dates.

huacnlee:
This is useful, but it not the primary target in our plan in currently.

If you really want it, you can make PR, we'd like to accept this improve.

PR #873, opened 2025-05-20
calendar: support custom matcher

0x1997:
See #870

### DIFF
--- a/crates/story/src/date_picker_story.rs
+++ b/crates/story/src/date_picker_story.rs
@@ -1,4 +1,4 @@
-use chrono::{Days, Duration, Utc};
+use chrono::{Datelike, Days, Duration, Utc};
 use gpui::{
     px, App, AppContext, Context, Entity, Focusable, IntoElement, ParentElement as _, Render,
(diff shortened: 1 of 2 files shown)


huacnlee:
Would you please add this as example to `CalendarStory`.

crates/story/src/calendar_story.rs
```diff
@@ -1,70 +1,225 @@
...
+                section("Large").max_w_md().child(
+                    DatePicker::new(&self.date_picker_large)
+                        .large()
+                        .width(px(300.)),
+                ),
+            )
+            .child(
+                section("Custom (First 5 days of each month disabled)")
                     .max_w_md()
-                    .child(Calendar::new(&self.calendar_wide).number_of_months(3)),
```
huacnlee: You are wrong, there have a date picker story, you don't need modify this file.

huacnlee: https://github.com/longbridge/gpui-component/blob/main/crates/story/src/date_picker_story.rs

huacnlee:
Thank you.

PR #883, opened 2025-05-22
feat: Add Postman-like story for making HTTP requests

xydz:
This commit introduces a new 'PostmanStory' to the storybook, providing a UI for you to make HTTP requests and view their responses.

Key features include:
- URL input and HTTP method selection (GET, POST, PUT, DELETE, etc.) using a dropdown.
- Tabbed interface for configuring:
    - Request Headers: Dynamically add/remove key-value pairs.
    - Query Parameters: Dynamically add/remove key-value pairs.
    - Request Body: Text area for raw body input.
- Asynchronous HTTP request sending using `reqwest`.
- Display area for response:
    - Status code.
    - Response headers (scrollable, line-by-line).
    - Response body (scrollable, plain text).
- Loading state indication while requests are in progress.
- Error message display for issues during request building, sending, or response processing.
- Basic unit tests for the HttpMethod enum.
- Integration into the existing storybook gallery.

The PostmanStory allows you to test HTTP endpoints directly from within the application's story environment, similar to standalone tools like Postman.

### DIFF
--- a/crates/story/Cargo.toml
+++ b/crates/story/Cargo.toml
@@ -13,6 +13,7 @@ gpui-component = { workspace = true, features = ["webview"] }
 rand = "0.8"
 raw-window-handle = { version = "0.6", features = ["std"] }
 regex = "1"
+reqwest = { version = "0.11", features = ["json", "rustls-tls"], default-features = false }
 reqwest_client = { git = "https://github.com/huacnlee/zed.git", branch = "webview" }
 rust-embed = "8.5.0"
 serde = "1"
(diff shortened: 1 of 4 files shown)


huacnlee:
It looks good, but all the code by AI is not acceptable, you should modify the example to fix code style, tidy the UI details and complete the features.

And this `postman` example should move to `examples` folder to as a single example like the `code-editor.rs`.

I have just fixed some syntax error, and also fixed the tokio error to let it works with GPUI, now example can works.

PR #911, opened 2025-06-02
highlighter: Improve performance for large file.

huacnlee:


huacnlee:
Active line for wrapped lines:

image

PR #945, opened 2025-06-11
chore: Update GPUI and fix uniform_list API changes.

### LINKED ISSUE #943: Upstream breaking change cause compile error
https://github.com/zed-industries/zed/pull/32480 which change `uniform_list()` arguments length will cause compile error

This problem occurs with an empty project which w/o cargo lockfile

Image

enpitsuLin:
close #943

I am not sure about title specification of PR. pls feel free to edit.


huacnlee:
Thanks

PR #957, opened 2025-06-13
number_input: Allow signed numbers

victorhqc:
Allow negative numbers and explicit positive numbers.

Added these tests:

1. Only one sign at the start of the number.
2. No signs in the middle of the number.
3. No signs in the fractional part.
4. Sign doesn't break formatting.

https://github.com/user-attachments/assets/89c4f4c7-d3bf-4cc1-aeca-7e9bedd02e20

### DIFF
--- a/crates/ui/src/input/mask_pattern.rs
+++ b/crates/ui/src/input/mask_pattern.rs
@@ -202,10 +202,21 @@ impl MaskPattern {
                     return false;
                 }
 
+                // Only one symbol is valid
+                if int_part
+                    .chars()
+                    .filter(|ch| is_digit_sign(*ch))
+                    .collect::<Vec<_>>()
+                    .len()
+                    > 1
+                {
+                    return false;
+                }
+
                 // check if the integer part is valid
                 if !int_part
                     .chars()
-                    .all(|ch| ch.is_ascii_digit() || Some(ch) == *separator)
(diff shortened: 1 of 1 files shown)


crates/ui/src/input/mask_pattern.rs
```diff
@@ -202,10 +202,20 @@ impl MaskPattern {
...
 
+                // only one sign is valid
+                if int_part.chars().filter(is_sign).collect::<Vec<_>>().len() > 1 {
+                    return false;
+                }
+
+                // sign is only valid at the beginning of the string
+                if int_part.chars().position(|ch| is_sign(&ch)) > Some(0) {
+                    return false;
+                }
```
huacnlee, line 218: It's looks like the second condition can cover the previous one?

crates/ui/src/input/mask_pattern.rs
```diff
@@ -379,6 +404,10 @@ impl MaskPattern {
     }
 }
 
+fn is_sign(ch: &char) -> bool {
```
huacnlee, line 412: Add `#[inline]` here

crates/ui/src/input/mask_pattern.rs
victorhqc: I tried that, but it broke the case of someone writing `+-`. I changed the code to avoid iterating over the characters twice and using one `if condition`. wdyt?

huacnlee:
Thank you.

PR #1019, opened 2025-06-29
sidebar: Add to support ContextMenu for SidebarMenuItem.

hlcfan:
Support for a context menu in the SidebarMenuItem component. My use case is that I want to show a context menu instead of only allowing to clicking on the menu item.

Please advise if this is a proper approach.

image

### DIFF
--- a/crates/ui/src/sidebar/menu.rs
+++ b/crates/ui/src/sidebar/menu.rs
@@ -64,14 +64,14 @@ pub struct SidebarMenuItem {
     handler: Rc<dyn Fn(&ClickEvent, &mut Window, &mut App)>,
     active: bool,
     collapsed: bool,
-    children: Vec<Self>,
-    context_menu: Vec<AnyElement>,
+    items: Vec<Self>,
+    children: Vec<AnyElement>,
     suffix: Option<AnyElement>,
 }
 
 impl ParentElement for SidebarMenuItem {
     fn extend(&mut self, elements: impl IntoIterator<Item = AnyElement>) {
-        self.context_menu.extend(elements);
+        self.children.extend(elements);
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/ui/src/sidebar/menu.rs
```diff
@@ -64,9 +65,18 @@ pub struct SidebarMenuItem {
     active: bool,
     collapsed: bool,
     children: Vec<Self>,
+    context_menu: Vec<AnyElement>,
     suffix: Option<AnyElement>,
 }
 
+impl ParentElement for SidebarMenuItem {
+    fn extend(&mut self, elements: impl IntoIterator<Item = AnyElement>) {
+        self.context_menu.extend(elements);
```
huacnlee: This is not correct, `ParentElement` should add to children, not a context_menu.

We should rename `context_menu` to `children`.

hlcfan: `SidebarMenuItem` has `children` field https://github.com/longbridge/gpui-component/blob/v20250617/crates/ui/src/sidebar/menu.rs#L66, how about renaming to `context_menu_children`

huacnlee: No, that is the wrong way!

ParentElement must to extend `children`, not any others.

huacnlee: I think, we can rename the old `children` to `items`, and add `children: Vec<AnyElement>`.

hlcfan: Updated. Can you share why it has to be `children`?

huacnlee:
I think this can't be just added feature to, there have some details need to do, when we right-clicked on an item, that item should have an active style.

Because of the Sidebar and SidebarMenuItem is an element, it is hard to hold a right_clicked state. We need to refactor the Sidebar design to use low-API with `with_element_state` to hold the right_clicked state.

Issue #851, opened 2025-05-12
In Scrollable example, the view on the bottom has no scrollbar (`.scrollable()`)


stippi:
Hi. I saw #478, but I'm not sure if it means the same issue. In my project, I am trying to convert the UI code to use `gpui-component`, but I seem to be having trouble getting scrollbars to work. I tried to understand the example. It seems like it shows two approaches to scrollbars, one using manual definition of the scrollable area with fine-grained control over the scrollbars. And (I'm not sure) the view on the bottom should show a simple approach based on overflow and computed layout sizes? If yes, it's what I need. But just like in my app, I see no scrollbar in the example.

huacnlee:
You are better to use `Scrollbar` in manually, like this example in:

https://github.com/longbridge/gpui-component/blob/main/crates/ui/src/list/list.rs#L280

stippi:
Thanks. In my project, the view I want to scroll has dynamically sized children. I rely on the layout system for the final size. In your example, I don't see where how needed size is computed, does it also rely on the layout system? I might need more guidance to fully understand what I need to do.

huacnlee:
Sorry, I don't know what do you want to do, you need explain more to details.

And the [scrollable_story.rs](https://github.com/longbridge/gpui-component/blob/main/crates/story/src/scrollable_story.rs) is the dynamical size, you can read and try to understand it.

Here is cases that used Scrollbar:

- https://github.com/longbridge/gpui-component/blob/1535df796e56aff900888f78fb8f48113d9e368c/crates/ui/src/dock/tiles.rs#L1144
- https://github.com/longbridge/gpui-component/blob/1535df796e56aff900888f78fb8f48113d9e368c/crates/ui/src/sidebar/mod.rs#L218
- https://github.com/longbridge/gpui-component/blob/1535df796e56aff900888f78fb8f48113d9e368c/crates/story/src/scrollable_story.rs#L277

stippi:
Ok, I'll describe it in non gpui related terms: I have these parts:
- **View 1** with lot's of different children. The view has a certain width. The children are laid out according to the width constraint and it results in an overall height for View 1 to accommodate all children.
- View 1 has a **parent** that has the same width as View 1, but also an independent height through some other constraint. That means view 1 is overflowing the parent.
- Now I want a scrollbar displaying within the parent on top of view 1, that is sized according to the height difference of view 1 and the parent, such that I can scroll the visible portion of view 1 within the parent.

[...]

PR #1042, opened 2025-07-04
fix the #851 issue:In Scrollable example, the view on the bottom has …

xudesheng:
…no scrollbar (.scrollable()) #851

```
    scroll_state: ScrollbarState,
    scroll_handle_list: ScrollHandle,
```

I have added a state to handle the scroll bar status for the list view. This code is entirely generated by `Claude Code`, and I have thoroughly validated it.

If I reuse the original `scroll_state`, the horizontal scroll on the first part will be impacted.

### DIFF
--- a/crates/story/src/scrollable_story.rs
+++ b/crates/story/src/scrollable_story.rs
@@ -323,7 +323,7 @@ impl Render for ScrollableStory {
                             .bottom_0()
                             .child(
                                 Scrollbar::both(&self.scroll_state_list, &self.scroll_handle_list)
-                                    .axis(ScrollbarAxis::Vertical),
+                                    .axis(self.axis),
                             )
                     })
             })
--- a/crates/story/src/scrollable_story.rs
+++ b/crates/story/src/scrollable_story.rs
@@ -1,7 +1,7 @@
 use std::rc::Rc;
 
 use gpui::{
-    div, px, size, App, AppContext, Axis, Context, Entity, Focusable, InteractiveElement,
+    div, px, size, App, AppContext, StatefulInteractiveElement, Context, Entity, Focusable, InteractiveElement,
     IntoElement, ParentElement, Pixels, Render, ScrollHandle, SharedString, Size, Styled, Window,
 };
 use gpui_component::{
@@ -19,6 +19,10 @@ pub struct ScrollableStory {
     scroll_handle: ScrollHandle,
     scroll_size: gpui::Size<Pixels>,
     scroll_state: ScrollbarState,
+    // handle the scrollable in the bottom list part, vertical scrollable only
+    scroll_handle_list: ScrollHandle,
+    scroll_state_list: ScrollbarState,
+
     items: Vec<String>,
     item_sizes: Rc<Vec<Size<Pixels>>>,
     test_width: Pixels,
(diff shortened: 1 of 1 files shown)


huacnlee:
I find any difference, please submit before, after screenshot or video for it.

crates/story/src/scrollable_story.rs
```diff
@@ -285,14 +291,17 @@ impl Render for ScrollableStory {
...
-                    .max_h(px(400.))
+                    // .max_h(px(400.)) // move the max_h to the next level
                     .min_h(px(200.))
                     .child(
                         v_flex()
                             .p_3()
                             .w(self.test_width)
                             .id("test-1")
-                            .scrollable(Axis::Vertical)
+                            // .scrollable(Axis::Vertical)
```
huacnlee, line 301: This is not needed to change, that container only have vertical scrollbar is by design, it not a bug.

xudesheng:
https://github.com/user-attachments/assets/5fbc90fe-cf2d-41d7-91d2-6d383fe3d53a


https://github.com/user-attachments/assets/57958862-3b4a-452b-a140-41c81a57d371

Please see the screen video recording before and after

crates/story/src/scrollable_story.rs
xudesheng: https://github.com/user-attachments/assets/645c962f-7958-4df7-8c23-60861f09c466

Got it, I changed it to follow the "self-axis" setting. When the setting is "Horizontal", you can scroll up/down only, and the vertical scroll bar is invisible, but the horizontal scroll bar is visible. I'm not sure if this will cause confusion or not.

huacnlee:
I see the bug it is only appeared in `cargo run --example scrollable`.

But, your changes are going to a wrong way, here I was used `scrollable` method, because I want to test this method. The `scrollable` is a special feature to extend any element to have a scrollbar. So we can't be replaced it by use `Scrollbar` directly.

The first time can't to scroll issue may be caused by `scrollable` method internal.

huacnlee (changes requested):
I see, the bug it is only appeared in `cargo run --example scrollable`.

But your changes are going to a wrong way. 

Here I was used `scrollable` method, because I want to test this method. The `scrollable` is a special feature to extend any element to have a scrollbar. So we can't be replaced it by use `Scrollbar` directly.

The first time can't to scroll issue may be caused by `scrollable` method internal, so we need to check it not change the example code.

xudesheng:
> I see, the bug it is only appeared in `cargo run --example scrollable`.
> 
> But your changes are going to a wrong way.
> 
> Here I was used `scrollable` method, because I want to test this method. The `scrollable` is a special feature to extend any element to have a scrollbar. So we can't be replaced it by use `Scrollbar` directly.
> 
> The first time can't to scroll issue may be caused by `scrollable` method internal, so we need to check it not change the example code.

Got your point and fully understood. It makes a lot of sense.

Now that I've given up on this fix, I'm looking forward to seeing your enhancement. Here is the response from Claude: 
我看到问题了 - 当前代码仍然在使用.scrollable(Axis::Vertical)，但这确实是一个复杂的布局问题。

你说得对，这个问题比我之前分析的复杂。.scrollable()方法在处理外部容器的max_h约束时确实存在根本性的架构问题。这种嵌套的约束传递在复杂的布局场景中很难完美解决。

如果你已经决定放弃这个问题，我完全理解。有时候某些UI架构的限制确实很难通过简单的修复来解决，特别是当涉及到复杂的布局约束传递时。

PR #1072, opened 2025-07-18
table: Add to support col_span, row_span.

huacnlee:


huacnlee:
The `col_span` and `row_span` is more complex, this may not a good idea to includes in to the Table.

This PR to just continue the Table API refactor.

0x1997:
`Column.align` is not respected

PR #1086, opened 2025-07-23
scrollbar: Fix content size

unixzii:
Caused by PR #1078 incorrect changes, and GPUI was changed API https://github.com/zed-industries/zed/pull/34832

The content size of the scrollable area should be its max scroll offset plus the container size.


huacnlee:
Thank you.

PR #1104, opened 2025-07-29
scrollbar: Reduce unnecessary frame updates

unixzii:
The current scrollbar implementation calls `Window::request_animation_frame` during the delay phase, which causes unnecessary CPU usage. This PR updates the scrollbar to use a timer to wait before starting the fade-out animation, allowing the CPU to remain idle during this phase.

There is the comparison with Instruments.
**Before:**
image

**After:**
image


huacnlee:
Thank you.

PR #1171, opened 2025-08-23
markdown: Fix that style attribute values are extracted correctly from img element

RemcoSmitsDev:
This PR fixes an issue with parsing the styles attribute values. The main issue it fixes is that we tried splitting on `:` twice which we should only do once.

Before this change see the following variable values: 
decl="width: 10px" 
rule=["width", "10px"] (each iteration one value)

And for each rule value we tried splitting again on `:` which is wrong we should only do this once.


huacnlee:
Thanks

PR #1214, opened 2025-09-07
input: Use zed's Rope

huacnlee:


huacnlee:
Still need to improvement:

- [ ] Use `shape_line` to instead of the `shape_text`, to only shape the visible lines.
- [ ] Improve the TextWrapper to only update the changed lines.

Issue #1220, opened 2025-09-08
input: Improve the large text performance.


huacnlee:
Still need to improvement:

- [x] Use `shape_line` to instead of the `shape_text`, to only shape the visible lines. #1222
- [x] Increment update syntax highlight #1230
- [x] Improve the TextWrapper to only update the changed lines. #1237
- [x] Avoid emit changed value in `InputEvent::Change`, this will spend 7% of time at every time editing.

<img width="1500" height="419" alt="Image" src="https://github.com/user-attachments/assets/2e5e1fd5-b8f4-40fb-9fac-9407843d50d2" />

huacnlee:
Completed improvement.

Now we can support 200K lines editing, and more than 500K lines read-only display.

PR #1222, opened 2025-09-08
input: Improve Input performance by only shape visible lines.

huacnlee:
Close #1193

Ref issue #1220 

This PR aim to shape text in visible ranges to reduce `shape_text` time for large document case.

- Fix rendering might be delayed and cause blanks when scrolling to follow the selection.
- Fix `RopeExt::lines` method to avoid call `line` method on skipped iter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /target
 .DS_Store
 /docks.json
+/profile.json
 .vscode
 index.scip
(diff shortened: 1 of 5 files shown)


huacnlee:
TODO: index_for_mouse_position need consider that the `lines` only have visible lines.

PR #1317, opened 2025-09-30
highlighter: Use clip_offset to ensure UTF-8 char boundaries

xudesheng:
## Problem

SyntaxHighlighter panics when highlighting text containing multi-byte UTF-8 characters (Chinese, Japanese, emoji, etc).

**Error**: `thread 'main' panicked at ropey-2.0.0-beta.1/src/rope.rs:694:59`

**Root Cause**: tree-sitter may return byte offsets that fall in the middle of multi-byte UTF-8 characters. When `Rope::slice()` receives such offsets, it checks `is_char_boundary()` and panics with `NonCharBoundary` error.

## Solution

Use `RopeExt::clip_offset()` to adjust byte offsets to the nearest char boundaries before calling `slice()`.

**Changes**:
- Line 395-398: Clip offsets in main highlighting loop
- Line 441-443: Clip offsets when extracting injection content  
- Line 478-479: Added explanatory comment for injection highlighting
- Added imports for `RopeExt` and `Bias`

**Safety**:
- `Bias::Left` for start offset ensures we don't skip the beginning of a character
- `Bias::Right` for end offset ensures we include the full character
- The adjustment is minimal (at most 3 bytes for UTF-8)
- Uses existing project API (`RopeExt::clip_offset`)

## Testing

Tested with:
- ✅ Chinese text: "你好世界"、"**加粗中文**"
- ✅ Japanese text: "こんにちは"、"日本語"
- ✅ Emoji: "😀🎉✨🚀"
- ✅ Markdown syntax highlighting with CJK characters
- ✅ Code blocks with mixed languages

All tests pass without panics. Syntax highlighting works correctly. No performance degradation observed.

## AI Assistance

🤖 This fix was developed with AI assistance (Claude). The solution approach (using `clip_offset`) was identified through code analysis and testing. The AI analyzed:
- ropey source code to understand the panic condition
- Existing usage of `clip_offset` in the codebase
- Tree-sitter byte offset behavior with UTF-8

All code has been reviewed and tested by humans, and further validated by Gemini AI.

## Checklist

- [x] Follows existing code style
- [x] One PR does one thing (UTF-8 panic fix only)
- [x] All manual tests pass
- [x] Tested with real multi-byte UTF-8 content
- [x] No breaking changes
- [x] No performance regression

### DIFF
--- a/crates/ui/src/highlighter/highlighter.rs
+++ b/crates/ui/src/highlighter/highlighter.rs
@@ -1,4 +1,5 @@
 use crate::highlighter::{HighlightTheme, LanguageRegistry};
+use crate::input::RopeExt;
 
 use anyhow::{anyhow, Context, Result};
 use gpui::{HighlightStyle, SharedString};
@@ -9,6 +10,7 @@ use std::{
     ops::Range,
     usize,
 };
+use sum_tree::Bias;
 use tree_sitter::{
     InputEdit, Node, Parser, Point, Query, QueryCursor, QueryMatch, StreamingIterator, Tree,
 };
@@ -433,14 +435,16 @@ impl SyntaxHighlighter {
         injection_language: &str,
         node: Node,
     ) -> Vec<(Range<usize>, String)> {
-        let start_offset = node.start_byte();
-        let end_offset = node.end_byte();
+        // Ensure byte offsets are on char boundaries for UTF-8 safety
+        let start_offset = self.text.clip_offset(node.start_byte(), Bias::Left);
+        let end_offset = self.text.clip_offset(node.end_byte(), Bias::Right);
+
         let mut cache = vec![];
         let Some(query) = &self.injection_queries.get(injection_language) else {
             return cache;
         };
 
-        let content = self.text.slice(node.byte_range());
+        let content = self.text.slice(start_offset..end_offset);
         if content.len() == 0 {
             return cache;
         };


crates/ui/src/highlighter/highlighter.rs
```diff
@@ -390,7 +392,10 @@ impl SyntaxHighlighter {
                     continue;
                 };
 
-                let node_range: Range<usize> = node.start_byte()..node.end_byte();
+                // Ensure byte offsets are on char boundaries for UTF-8 safety
+                let start_byte = self.text.clip_offset(node.start_byte(), Bias::Left);
+                let end_byte = self.text.clip_offset(node.end_byte(), Bias::Right);
+                let node_range: Range<usize> = start_byte..end_byte;
```
huacnlee: We don't need change here.

xudesheng: I'm not sure I get your point. 

```
// Current code (line 102-104)
let node_range: Range<usize> = node.start_byte()..node.end_byte();
let text = self.text.slice(node_range.clone());  // ❌ Panics here
```

tree-sitter returns **byte offsets**, but Rope requires **character boundary offsets**.

For Chinese text `"这是**粗体**"`:
- tree-sitter might return `start_byte = 7` 
- But byte 7 is **inside** the 3-byte UTF-8 sequence for `'是'` (bytes 6-8)
- Rope::slice() panics because 7 is not at a character boundary

huacnlee: I mean, this part not going to use `slice` method. So we do not need to convert the byte offset.

Only the `handle_injection` method need it.

xudesheng: You're right about `match_styles` - that part doesn't use `slice()` so it doesn't need the fix there.

**However, the real problem is in the `handle_injection` method** (not shown in my PR diff):
```
fn handle_injection(&self, injection_language: &str, node: Node) -> Vec<(Range<usize>, String)> {
    // ...
    let content = self.text.slice(node.byte_range());  // ❌ This line panics with UTF-8
    // ...
}
```

Sorry for the confusion - I was fixing the wrong location. Would you like me to move the fix to 'handle_injection' instead, or use a different approach?

huacnlee: Yes, you just need to change this `handle_injection`.

xudesheng: I've updated the PR to only fix `handle_injection` where `slice()` is actually called. The unnecessary change in `match_styles` has been removed. Force-pushed the corrected version. Thank you! 🙏

huacnlee:
Thank you.

PR #1318, opened 2025-10-01
feat: assets interpolate folder path via rust embed

carlocorradini:
Use `rust-embed` feature `interpolate-folder-path` to always (`debug`/`release`) load assets from the correct `assets` directory

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5864,6 +5864,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "rust-embed-utils",
+ "shellexpand",
  "syn 2.0.105",
  "walkdir",
 ]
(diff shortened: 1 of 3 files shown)


themonstergame:
So, what the problem do you meet before this change.

huacnlee:
So, what the problem do you meet before this change.

carlocorradini:
@huacnlee 
Hey, no issues on my side.
This PR is just a code simplification. Since asset loading is already handled by `rust-embed`, we don’t need to duplicate that logic (load via file system). With the `interpolate-folder-path` feature, we can directly use the `CARGO_MANIFEST_DIR` environment variable, which always resolves to an absolute folder path. This ensures the same assets are loaded in both debug and release modes.
Moreover, since we’re only loading `.svg` icons, I’ve added the extension to the `include` pattern. This makes the configuration more specific and removes the need for an exclude rule (e.g., to filter out files like `.DS_Store`).

huacnlee:
I don't think this change is necessary. 

The rust-embed to embed the files at the compile time, the `../../assets` are same for debug or other any build target. 

The original way is correct, we do not need to change it.

carlocorradini:
@huacnlee 
From [rust-embed](https://crates.io/crates/rust-embed):
> The path resolution works as follows:
>    In `debug` and when `debug-embed` feature is not enabled, the folder path is resolved relative to where the binary is run from.
>    In `release` or when `debug-embed` feature is enabled, the folder path is resolved relative to where `Cargo.toml` is.

This PR simply removes unnecessary code (filesystem asset loading), as `rust-embed` already handles this directly. Fewer lines of code to maintain.

huacnlee:
Oh, you are right. 

Now I know this difference.

huacnlee:
Thank you for explain.

PR #1337, opened 2025-10-08
context_menu: Fix subscription leak.

huacnlee:


ghost:
Hi @huacnlee just a question.

I wonder why do we have to do `_subscription: `, it make the code looks quite clunky.

From the [documentation](https://github.com/zed-industries/zed/blob/5c194f7cdc57fdbb93246f4623db9ad471d8511e/crates/gpui/src/subscription.rs#L172-L174), can we just detach it then when the entity dropped the subscription will be dropped.

Maybe there are some technicalities?

huacnlee:
Use detach may have memory leaks.

So, the subscription is better to manage by the owner. When the owner is going to drop. The subscription will ensure to drop.

ghost:
Yeah I understand the intention here, I just wonder if we can repro it somehow so that we can find a way to make the code better.

PR #1356, opened 2025-10-12
plot: Add StepAfter kind to StokeStrike

itsmgo:
Implements new `StokeStrike::StepAfter`  for plot component

step-after-plot-stoke-style


madcodelife:
Thank you!

PR #1364, opened 2025-10-13
plot: Add stroke style per each Y series

itsmgo:
Enables the user to set different `stroke_style` to each Y data series, keeping retrocompatibility with the previous API.

Screenshot from 2025-10-13 11-33-28

### DIFF
--- a/crates/ui/src/chart/area_chart.rs
+++ b/crates/ui/src/chart/area_chart.rs
@@ -23,9 +23,9 @@ where
     data: Vec<T>,
     x: Option<Rc<dyn Fn(&T) -> X>>,
     y: Vec<Rc<dyn Fn(&T) -> Y>>,
-    stroke: Vec<Hsla>,
-    stroke_style: Vec<StrokeStyle>,
-    fill: Vec<Background>,
+    strokes: Vec<Hsla>,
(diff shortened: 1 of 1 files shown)


crates/ui/src/chart/area_chart.rs
```diff
@@ -24,7 +24,7 @@ where
     x: Option<Rc<dyn Fn(&T) -> X>>,
     y: Vec<Rc<dyn Fn(&T) -> Y>>,
     stroke: Vec<Hsla>,
-    stroke_style: StrokeStyle,
+    stroke_style: Vec<StrokeStyle>,
```
huacnlee: stroke_styles

crates/ui/src/chart/area_chart.rs
```diff
@@ -24,7 +24,7 @@ where
     x: Option<Rc<dyn Fn(&T) -> X>>,
     y: Vec<Rc<dyn Fn(&T) -> Y>>,
     stroke: Vec<Hsla>,
```
huacnlee: Please also rename this too.

itsmgo: I also took the opportunity to rename `fill` to `fills`

madcodelife:
Thank you!

PR #1370, opened 2025-10-14
story: Add interactive brush drawing canvas

han1548772930:
image


han1548772930:
> Thank you.
> 
> But would you please move this to `examples/` folder to as a single example? Because this is not a component of the GPUI Component, I think it should not be a store in the gallery.

Okay, no problem.

PR #1404, opened 2025-10-21
chore: Update dependencies to smol = 2

vincenthz:
just noticed the dependency is set to a fairly old version of smol (= 1), so update to smol = 2.

Also gpui/zed seems to be using smol = 2


huacnlee:
Thanks

sjmackenzie:
Great to see you guys are responsive. Kudos, for the work into this project.

PR #1411, opened 2025-10-22
table: Refactor Table to use callback instead of delegate API.

huacnlee:
## Break Changes

- `Table` changed to stateless mode to implement `RenderOnce`.
- `TableDelegate` has been removed.

```rs
struct MyView {
  table: Entity<TableState>,
}

impl MyView {
    fn new(window: &mut Window, cx: &mut App) -> Self {
        let columns = vec![
            Column::new("name", "Name").width(120.)
            Column::new("age", "Age").width(60.)
        ];

        let table = cx.new(|cx| TableState::new(columns, 100));
        Self {
            table,
        }
    }
}

impl Render for MyView {
    fn render(&mut self, cx: &mut App, window: &mut Window, cx: &mut Context<Self>) -> IntoElement {
        Table::new(self.table)
            .small()
            .stripe(true)
            .cell(cx.processor(|this, (row_ix, col_ix), window, cx| {
                div().child(format!("Cell {}, {}", row_ix, col_ix)))
            }))
}
```

### DIFF
--- a/docs/docs/components/table.md
+++ b/docs/docs/components/table.md
@@ -10,63 +10,66 @@ A comprehensive data table component designed for handling large datasets with h
 ## Import
 
 ```rust
-use gpui_component::table::{Table, TableDelegate, Column, ColumnSort, ColumnFixed, TableEvent};
+use gpui_component::table::{Table, TableState, Column, ColumnSort, ColumnFixed, TableEvent};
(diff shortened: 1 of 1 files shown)


huacnlee:
Close this, this change will make more complex in the callback function.

PR #1425, opened 2025-10-24
story: Fix image story not loaded svg

### LINKED ISSUE #1418: The sample image cannot be loaded
`2025-10-23T08:19:17.654004Z ERROR gpui::asset_cache: Failed to load asset: asset error: Embedded resource not found: src/fixtures/color-wheel.svg`

iainlau:
Close #1418


huacnlee:
image

SVG color incorrect.

huacnlee:
Here is a bit complex I think we can ignore.

You can just remove them or use other image from an online URL like this third one.

This 2 case is not need now, because I was removed `SvgImg` before (#1187), this 2 case is for test `SvgImg`.

iainlau:
Copy that.

PR #1438, opened 2025-10-28
Change gpui-component version to 0.3.1

scriptandcompile:
Updated gpui-component version in README. 

Removed verbage about requiring git based toml including since the crate is now released on crates.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,9 @@ We built multi-theme support in the application. This feature is not included in
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


huacnlee:
Thanks for make change.

This is not needed, the Cargo will use the newest version.

scriptandcompile:
It was mostly about the line suggesting to use git instead of the crates toml version include I was changing since it's now on crates.io.

PR #1440, opened 2025-10-28
title_bar: fix drag behaviour when the left click is hold while entering the area

Dadoum:


Dadoum:
It is marked as a draft since I don't really know what is the proper way of adding state to the element here.

```
.child(TitleBarElement { state: cx.new(|_| TitleBarElementState { can_start_move: false }) }),
```

is obviously wrong, I know.

PR #1466, opened 2025-10-30
linux: Disallow resizing window when maximized

Dadoum:
A simple example is that of maximized windows. When maximized, the window can still be resized as its edges. Furthermore, this commit also adds a bigger corner hit box.


huacnlee:
Please upload video for show before vs after about the issue.

Dadoum:
After testing all borders it seems that I can still trigger the bug on the right edge! Here is a video of that bug:

madcodelife:
Closes on #2052.

PR #1475, opened 2025-11-02
Replace Input with TextInput in main.rs

wuqiong4945:


huacnlee:
This is wrong. We have renamed it to Input.

PR #1478, opened 2025-11-03
modal: Block background interaction when Modal is active.

FlyingYu-Z:
### Before
The ContextMenu could still open behind the modal.

<video src="https://github.com/user-attachments/assets/8358e7dc-1a59-4053-90ad-9b28723a4555" controls loop muted style="max-width: 100%; height: auto;"></video>

### After
Background ContextMenu is blocked when overlay(true).

<video src="https://github.com/user-attachments/assets/c98ef3b7-88ad-415e-9a03-4b04eb9befe7" controls loop muted style="max-width: 100%; height: auto;"></video>

### DIFF
--- a/crates/story/src/drawer_story.rs
+++ b/crates/story/src/drawer_story.rs
@@ -15,7 +15,7 @@ use gpui_component::{
     date_picker::{DatePicker, DatePickerState},
     h_flex,
     input::{Input, InputState},
-    list::{List, ListDelegate, ListItem},
+    list::{ListDelegate, ListItem, ListState},
     v_flex,
     webview::WebView,
     wry,
@@ -43,7 +43,7 @@ impl ListDelegate for ListItemDeletegate {
         &mut self,
         query: &str,
         _: &mut Window,
-        cx: &mut Context<List<Self>>,
+        cx: &mut Context<ListState<Self>>,
     ) -> Task<()> {
         let query = query.to_string();
         cx.spawn(async move |this, cx| {
@@ -65,12 +65,7 @@ impl ListDelegate for ListItemDeletegate {
         })
     }
 
-    fn render_item(
-        &self,
-        ix: IndexPath,
-        _: &mut Window,
-        _: &mut Context<List<Self>>,
-    ) -> Option<Self::Item> {
+    fn render_item(&self, ix: IndexPath, _: &mut Window, _: &mut App) -> Option<Self::Item> {
         let confirmed = Some(ix.row) == self.confirmed_index;
 
         if let Some(item) = self.matches.get(ix.row) {
@@ -102,7 +97,7 @@ impl ListDelegate for ListItemDeletegate {
         }
     }
 
-    fn render_empty(&self, _: &mut Window, cx: &mut Context<List<Self>>) -> impl IntoElement {
+    fn render_empty(&self, _: &mut Window, cx: &mut App) -> impl IntoElement {
         v_flex()
             .size_full()
             .child(
@@ -118,13 +113,13 @@ impl ListDelegate for ListItemDeletegate {
             .text_color(cx.theme().muted_foreground)
     }
 
-    fn cancel(&mut self, window: &mut Window, cx: &mut Context<List<Self>>) {
+    fn cancel(&mut self, window: &mut Window, cx: &mut Context<ListState<Self>>) {
         _ = self.story.update(cx, |this, cx| {
             this.close_drawer(window, cx);
         });
     }
 
-    fn confirm(&mut self, _secondary: bool, _: &mut Window, cx: &mut Context<List<Self>>) {
+    fn confirm(&mut self, _secondary: bool, _: &mut Window, cx: &mut Context<ListState<Self>>) {
         _ = self.story.update(cx, |this, _| {
             self.confirmed_index = self.selected_index;
             if let Some(ix) = self.confirmed_index {
@@ -139,7 +134,7 @@ impl ListDelegate for ListItemDeletegate {
         &mut self,
         ix: Option<IndexPath>,
         _: &mut Window,
-        cx: &mut Context<List<Self>>,
+        cx: &mut Context<ListState<Self>>,
     ) {
         self.selected_index = ix.map(|ix| ix.row);
 
@@ -153,7 +148,7 @@ pub struct DrawerStory {
     focus_handle: FocusHandle,
     drawer_placement: Option<Placement>,
     selected_value: Option<SharedString>,
-    list: Entity<List<ListItemDeletegate>>,
+    list: Entity<ListState<ListItemDeletegate>>,
     input1: Entity<InputState>,
     input2: Entity<InputState>,
     date: Entity<DatePickerState>,
@@ -248,7 +243,7 @@ impl DrawerStory {
             matches: items.clone(),
         };
         let list = cx.new(|cx| {
-            let mut list = List::new(delegate, window, cx);
+            let mut list = ListState::new(delegate, window, cx);
             list.focus(window, cx);
             if let Some(query_input) = list.query_input() {
                 query_input.update(cx, |input, cx| {
--- a/crates/story/src/list_story.rs
+++ b/crates/story/src/list_story.rs
@@ -2,18 +2,19 @@ use std::{rc::Rc, time::Duration};
 
 use fake::Fake;
 use gpui::{
-    actions, div, prelude::FluentBuilder as _, px, App, AppContext, Context, Edges, ElementId,
-    Entity, FocusHandle, Focusable, InteractiveElement, IntoElement, ParentElement, Render,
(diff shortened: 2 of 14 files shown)


huacnlee:
Please also to fix the drawer.

crates/ui/src/modal.rs
```diff
@@ -248,6 +248,9 @@ impl Modal {
     ///
     /// When the overlay is clicked, the modal will be closed.
     pub fn overlay_closable(mut self, overlay_closable: bool) -> Self {
+        if !self.overlay && overlay_closable {
+            eprintln!("Warning: overlay_closable has no effect when overlay is disabled.");
+        }
```
huacnlee: Keep 1 PR to 1 thing to help us to review and merge.

This change is not good enough, you can revert it.

crates/ui/src/modal.rs
```diff
@@ -382,25 +385,30 @@ impl RenderOnce for Modal {
...
+                                }
+                                // If overlay is true, then block background interaction, such as ContextMenu.
+                                // (for example, prevent right-click menus from opening behind the modal)
+                                cx.stop_propagation();
                             }
                         })
                     })
+                    .when(self.overlay_visible, |this| {
+                        this.occlude().bg(overlay_color(self.overlay, cx))
+                    })
```
huacnlee: This is looks like not change, right?

If you are not changed this, please keep the original source code ordering, to avoid new diff change.

FlyingYu-Z: Yes, I didn't change it.

huacnlee:
I have just did a test, there is no issue like the video of you submit.

https://github.com/user-attachments/assets/78e7a727-b905-4331-a2ae-4e9220367910

You can try the `editor` example:

```
cargo run --example editor
```

FlyingYu-Z:
> I have just did a test, there is no issue like the video of you submit.
> 
>  Untitled.mov 
> You can try the `editor` example:
> 
> ```
> cargo run --example editor
> ```

Because I didn't use the internal TreeView. I made my TreeView:
```diff
    fn class_node_view(
        &self,
        window: &mut Window,
        cx: &mut Context<Self>,
        node: &Arc<ClassTreeNode>,
    ) -> impl IntoElement {
        let node_for_event = Arc::clone(node);
        let name = node.name.clone();
        let level = node.level;
        let is_directory: bool = node.is_directory;
        let is_expanded = node.is_directory_expanded();
        let id = node.identifier;
        let indent = px(20.0 * (level - 1) as f32);
        let selected_class_id = self.selected_class_id;
        div()
            // .w(px(500.))
            .w_full()
            .child(
                div()
                    .id(ElementId::Integer(id))
                    .on_any_mouse_down(cx.listener(move |this, e, window, cx| {
                        this.selected_class_id = id;
                        cx.notify();
                    }))
                    // .when(!window.has_active_modal(cx), |container| {
                    //     container
+                           .context_menu({
+                              move |this, window, cx| {
+                                 this.separator()
+                                       .menu("Open", Box::new(Open))
+                                      .menu("Delete", Box::new(Delete))
+                                      .when(!is_directory, |this| {
+                                          this.menu("Export", Box::new(Export))
+                                              .menu("Info", Box::new(Info))
+                                      })
+                                      .separator()
+                              }
+                          })
                    // })
                    .when(selected_class_id == id, |container| {
                        container.bg(rgba(0x015f5fff)).cursor_pointer()
                    })
                    .on_double_click(cx.listener(move |this, e, window, cx| {
                        this.open_class(&node_for_event, window, cx);
                    }))
                    .flex()
                    .items_start()
                    .gap_1()
                    .pl(indent)
                    .when(is_directory, |container| {
                        container.when_else(is_expanded, 
                            |container| {
                        container.child(IconName::ChevronDown)
                        }, |container| {
                        container.child(IconName::ChevronRight)
                        })
                    })
                    .when_else(
                        is_directory,
                        |container| container.child(self.icon_folder()),
                        |container| container.child(self.icon_class()),
                    )
                    .child(Label::new(name)),
            )
            .when(
                is_directory && is_expanded,
                |container| {
                    let items: Vec<AnyElement> = node.get_children()
                                                .into_iter()
                                                .map(|child_node| self.class_node_view(window, cx, &child_node).into_any_element())
                                                .collect::<Vec<AnyElement>>();
                    container.children(items)
                },
            )
    }
```

FlyingYu-Z:
This is the  demo, you can check it.
```rust

use gpui::*;
use gpui_component::{button::*, label::Label, menu::ContextMenuExt, *};

actions!(class_menu, [Open, Delete, Export, Info]);

pub struct HelloWorld;

impl HelloWorld {
    
    fn show_test_modal(&mut self, _: &ClickEvent, window: &mut Window, cx: &mut Context<Self>) {
        window.open_modal(cx, move |modal, window, cx| {
            modal
                .title("Tset Modal")
                .overlay(true)
                .overlay_closable(true)
                .child(
                    div().w(px(500.0))
                    .h(px(300.0))
                    .bg(gpui::blue())
                )
                .confirm()
                .on_ok({
                    move |_, window, cx| {
                        true
                    }
                })
        });
    }
}

impl Render for HelloWorld {

    fn render(&mut self, window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
        let modal_layer = Root::render_modal_layer(window, cx);
        div()
            .v_flex()
            .gap_2()
            .size_full()
            .child(
                div().flex_1().pt_40().child(
                    Button::new("btn1")
                        .primary()
                        .label("Click me first")
                        .on_click(cx.listener(Self::show_test_modal)),
                ),
            )
            .child(
                div()
                    .id("second-area")
                    .v_flex()
                    .h_40()
                    .bg(gpui::yellow())
                    .items_center()
                    .justify_center()
                    .child(Label::new("Then click me with Mouse-Right"))
                    .context_menu({
                        move |this, window, cx| {
                            this.separator()
                                .menu("Open", Box::new(Open))
                                .menu("Delete", Box::new(Delete))
                                .menu("Export", Box::new(Export))
                                .menu("Info", Box::new(Info))
                                .separator()
                        }
                    }),
            )
            .children(modal_layer)
    }
}

fn main() {
    let app = Application::new();

    app.run(move |cx| {
        // This must be called before using any GPUI Component features.
        gpui_component::init(cx);

        cx.spawn(async move |cx| {
                let view = cx.new(|_| HelloWorld);
                // This first level on the window, should be a Root.
                cx.new(|cx| Root::new(view.into(), window, cx))
            })?;

            Ok::<_, anyhow::Error>(())
        })
        .detach();
    });
}
```

crates/ui/src/modal.rs
```diff
@@ -385,19 +387,24 @@ impl RenderOnce for Modal {
                     .when(self.overlay_visible, |this| {
                         this.occlude().bg(overlay_color(self.overlay, cx))
                     })
-                    .when(self.overlay_closable, |this| {
+                    .when(self.overlay, |this| {
```
huacnlee: This is not correct, we not need change here.

crates/ui/src/modal.rs
```diff
@@ -385,19 +387,24 @@ impl RenderOnce for Modal {
                     .when(self.overlay_visible, |this| {
                         this.occlude().bg(overlay_color(self.overlay, cx))
```
huacnlee: Try move `occlude` to out side of the condition. This method will blocking any mouse interaction.

crates/ui/src/drawer.rs
```diff
@@ -144,13 +144,15 @@ impl RenderOnce for Drawer {
                     .w(size.width)
                     .h(size.height - titlebar_height)
                     .bg(overlay_color(self.overlay, cx))
-                    .when(self.overlay_closable, |this| {
-                        this.on_mouse_down(MouseButton::Left, {
-                            let on_close = self.on_close.clone();
-                            move |_, window, cx| {
+                    .when(self.overlay, |this| {
```
huacnlee: So the drawer was correctly, we don't need to change.

crates/ui/src/modal.rs
FlyingYu-Z: Hi, I’d like to clarify the relationship between `overlay` and `overlay_closable`.

`overlay_closable` only makes sense if `overlay` is enabled. 
If `overlay` is set to `false`, there is no overlay layer rendered, 
so clicking the background cannot close the modal. In other words, 
`overlay_closable` has no effect without `overlay`.

crates/ui/src/drawer.rs
FlyingYu-Z: From a UI design perspective, `overlay_closable` should depend on `overlay`:

1. `overlay_closable` controls clicking the overlay to close the Modal. 
   If there is no overlay, there is nothing to click, so the property has no effect.
2. It maintains consistent user expectations: clicking the background either closes the Modal or does nothing visible.
3. Overlay provides visual feedback for an active Modal. Without it, the "click to close" interaction would be invisible and confusing.

Thus, overlay_closable logically requires overlay to be enabled.

huacnlee:
Sorry, I have made a mistake for git push.

PR #1497, opened 2025-11-04
input: Delete selected text with modifier keys

Moulberry:
In most applications, when text is selected and backspace/delete is pressed, the selection is simply deleted.

Currently, gpui-component will ignore the selected text and delete the previous/next word when Ctrl is pressed.

This can be very confusing especially if you do ctrl+a then ctrl+backspace, expecting the input to be cleared.

This PR makes it so that if there is an active selection, the active selection is deleted before doing any extra processing.

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -1062,19 +1062,20 @@ impl InputState {
     ) {
         if !self.selected_range.is_empty() {
             self.replace_text_in_range(None, "", window, cx);
-        } else {
-            let mut offset = self.start_of_line();
-            if offset == self.cursor() {
-                offset = offset.saturating_sub(1);
-            }
-            self.replace_text_in_range_silent(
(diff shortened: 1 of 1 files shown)


crates/ui/src/input/state.rs
```diff
@@ -1060,16 +1060,20 @@ impl InputState {
         window: &mut Window,
         cx: &mut Context<Self>,
     ) {
-        let mut offset = self.start_of_line();
-        if offset == self.cursor() {
-            offset = offset.saturating_sub(1);
+        if !self.selected_range.is_empty() {
+            self.replace_text_in_range(None, "", window, cx);
```
huacnlee, line 1064: return here to avoid touch the origin method.

Moulberry: [02195d6](https://github.com/longbridge/gpui-component/pull/1497/commits/02195d66e5641490a7fda88275daec0df357edb2)

huacnlee:
Thank you.

PR #1500, opened 2025-11-04
list: Fix selection when selectable(false) and select first element after search

Moulberry:
Fixes two issues.

1. Even when selectable(false) is used, previously the list could still be selected using the arrow keys or via searching
(If this is intended please let me know, but I think this is the wrong behaviour)

2. When searching, previously the selection would apply *before* the search completes.
Now, it will defer the selection of the element until after the search completes *and* the rows are updated.


huacnlee:
Sorry, I have made a better change in #1508

PR #1502, opened 2025-11-04
select/date_picker: Set default cleanable state to false

Moulberry:
The default was accidentally set to true for Select in 893e323.
The Select::cleanable() method sets it to true, so the default should be false.

(Same with DatePicker, I just searched for `cleanable: true` in the codebase and it had the same problem :) )

### DIFF
--- a/crates/ui/src/select.rs
+++ b/crates/ui/src/select.rs
@@ -306,7 +306,7 @@ impl Default for SelectOptions {
             style: StyleRefinement::default(),
             size: Size::default(),
             icon: None,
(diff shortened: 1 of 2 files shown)


huacnlee:
We are discussed of this, we'd like to keep `cleanable: true`.

Please change the `fn cleanable()` to `fn cleanable(mut self, cleanable: bool)` like the `disabled` method.

Including `Input`, `Select` and `DatePicker`, please search the project, if there have others like this, we need also to update it.

PR #1505, opened 2025-11-04
table: fix row border missing on first frame

Moulberry:
Currently, the bounds are zero on the first frame.
Therefore, the extra_rows_count will always be zero on the first frame.
This causes rendering issues for row border & row stripes.

Unfortunately, I don't know how to get the correct height on the first frame to fix the row stripes.

However, an easy fix for the row border rendering issue is to disable the table_is_filled check. The check is problematic anyways since the border affects the height of the element.

I attached a video showing the issue, the border doesn't render on the first frame and the height of the element also shifts due to the border:

https://github.com/user-attachments/assets/b154c686-cd9c-4b6b-8ba6-c79640297abc

### DIFF
--- a/crates/ui/src/table/mod.rs
+++ b/crates/ui/src/table/mod.rs
@@ -1006,7 +1006,7 @@ where
 
         if row_ix < rows_count {
             let is_last_row = row_ix + 1 == rows_count;
-            let need_render_border = !is_last_row || !is_filled;
+            let need_render_border = is_selected || !is_last_row || !is_filled;
 
             let mut tr = self.delegate.render_tr(row_ix, window, cx);
             let style = tr.style().clone();


huacnlee:
Your change is not correct, now the vertical scrollbar is overflow to appear.

image

Moulberry:
> Your change is not correct, now the vertical scrollbar is overflow to appear.

Yes, I noticed this too.
However, the stripe vertical scrollbar bug happens even without this change.

I took this screenshot on the current latest commit (without this pull request), and you can see the scrollbar is visible too:
image

Please check the story example on the latest commit to confirm the bug isn't caused by this change

huacnlee:
You can try my update, now main branch not overflow to show scrollbar.

But your changes are still having.

Moulberry:
I removed the change that was causing it.
I also removed the code for disabling rendering the last border, it seems wrong:

Before:
image
After:
image

huacnlee:
We need to avoid render last row bottom border.

image

Moulberry:
> We need to avoid render last row bottom border.

Ah, I understand the purpose now.

I have added a commit which makes it so the border isn't rendered when the table is full, but renders the border otherwise.

Is that correct?

Moulberry:
Hmm I think it also needs to render the border when selected

huacnlee:
Thanks

PR #1506, opened 2025-11-04
input, select, date_picker: Refactor `cleanable` to have argument and default to false.

Moulberry:
https://github.com/longbridge/gpui-component/pull/1502#issuecomment-3485354324

## Break Change

- The `DatePicker` and `Select` has changed `cleanable` default from `true` to `false`.
- Updated `Input`, `DatePicker` and `Select` the `cleanable` to have a argument.

```diff
- pub fn cleanable(mut self)
+ pub fn cleanable(mut self, cleanable: bool)
```

### DIFF
--- a/crates/story/src/main.rs
+++ b/crates/story/src/main.rs
@@ -220,7 +220,8 @@ impl Render for Gallery {
                                             .mx_1()
                                             .child(
                                                 Input::new(&self.search_input)
-                                                    .appearance(false),
+                                                    .appearance(false)
+                                                    .cleanable(true),
                                             ),
                                     ),
                             )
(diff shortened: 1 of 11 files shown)


crates/story/src/main.rs
```diff
@@ -220,8 +220,7 @@ impl Render for Gallery {
                                             .mx_1()
                                             .child(
                                                 Input::new(&self.search_input)
-                                                    .appearance(false)
-                                                    .cleanable(),
+                                                    .appearance(false),
```
huacnlee: Here should be `cleanable(true)`

huacnlee: Oh, it correct

huacnlee:
@Moulberry After some consideration, I realized that having all input fields set to cleanable by default isn't ideal, as it's not always necessary. Therefore, I've decided to revert it to the default value of false.

Moulberry:
Okay, I understand.

Moulberry:
Oh I didn't realize that you modified this pull request to make it false by default. My Bad

huacnlee:
I will continue to consider more of this change tonight, then later to merge.

PR #1537, opened 2025-11-07
button_group: Add vertical

Moulberry:
Adds a vertical option to the button group

Also added an example of the vertical button group to the story

Also fixed button rounding only working when both tl/bl and tr/br are both set, now they can be set individually

I'm not sure if the justify_center in `.when(self.vertical, |this| this.flex_col().justify_center())` is needed, but I added it to match the behaviour of `.items_center()` when the flex direction is row.

### DIFF
--- a/crates/story/src/button_story.rs
+++ b/crates/story/src/button_story.rs
@@ -1,5 +1,5 @@
 use gpui::{
-    Action, App, AppContext as _, ClickEvent, Context, Corner, Entity, Focusable,
+    Action, App, AppContext as _, Axis, ClickEvent, Context, Corner, Entity, Focusable,
     InteractiveElement, IntoElement, ParentElement as _, Render, Styled as _, Window,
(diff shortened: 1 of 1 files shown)


crates/ui/src/button/button_group.rs
```diff
@@ -18,6 +18,7 @@ pub struct ButtonGroup {
     children: Vec<Button>,
     pub(super) multiple: bool,
     pub(super) disabled: bool,
+    pub(super) vertical: bool,
```
huacnlee: Use `axis`, default to `Hozizontal`

Moulberry: [929020f](https://github.com/longbridge/gpui-component/pull/1537/commits/929020faa22520df80f42fce124ec8df4965e10b)

huacnlee:
Thank you.

PR #1542, opened 2025-11-08
slider: Add option to disable tooltip

CherryWorm:
Sometimes it's desirable to not display a tooltip, especially if the values might not make intuitive sense to a user or they're already displayed elsewhere

### DIFF
--- a/crates/ui/src/slider.rs
+++ b/crates/ui/src/slider.rs
@@ -263,6 +263,7 @@ pub struct Slider {
     axis: Axis,
     style: StyleRefinement,
     disabled: bool,
+    display_tooltip: bool,
 }
 
 impl Slider {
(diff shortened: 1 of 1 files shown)


crates/ui/src/slider.rs
```diff
@@ -282,6 +284,12 @@ impl Slider {
         self
     }
 
+    /// Set whether to display the tooltip on the thumb, default: true
+    pub fn display_tooltip(mut self, display_tooltip: bool) -> Self {
+        self.display_tooltip = display_tooltip;
+        self
+    }
```
huacnlee, line 291: This is not a good API.

If we want to have this option, we need to consider to choice a better one. 

We are better to find any reference like this feature in other UI frameworks.

crates/ui/src/slider.rs
```diff
@@ -350,6 +358,7 @@ impl Slider {
             .on_drag_move(window.listener_for(
                 &self.state,
                 move |view, e: &DragMoveEvent<DragThumb>, window, cx| {
+                    println!("Dragging: {:?}", e.event);
```
Moulberry: Leftover debugging that should not be included

CherryWorm: Sorry about that, fixed it

crates/ui/src/slider.rs
CherryWorm: What API would you prefer? I couldn't find any UI framework that displays the current slider value as a tooltip, so I also couldn't find a fitting api for disabling this.  I'd really like to have this option though, the current implementation isn't very good as it doesn't even properly round the values.

huacnlee:
I think, we can remove this tooltip, that is no useful. 

I have checked the slider on macOS, and shadcn/ui, they are both not has tooltip.

PR #1543, opened 2025-11-08
slider: Add option to choose between linear and logarithmic scale

CherryWorm:
Adds the ability to choose between a linear and a logarithmic scale for the slider. A logarithmic scale is the right and intuitive choice for many different slider applications. Building this right into the component has two advantages:
- The user doesn't have to convert at every point where they might use or update the slider value
- On a logarithmic scale, the distance between steps varies over the sliders range. This implementation respects that

### DIFF
--- a/crates/ui/src/slider.rs
+++ b/crates/ui/src/slider.rs
@@ -142,7 +142,9 @@ pub enum SliderScale {
     ///
     /// # For example
     ///
-    /// ```rust
+    /// ```
+    /// use gpui_component::slider::{SliderState, SliderScale};
+    ///
     /// let slider = SliderState::new()
     ///     .min(1.0)    // Must be > 0 for logarithmic scale
     ///     .max(1000.0)


crates/ui/src/slider.rs
```diff
@@ -192,16 +228,42 @@ impl SliderState {
...
+    /// converts a value between the minimum and maximum value to a value between 0.0 and 1.0, depending on the chosen scale
+    fn value_to_percentage(&self, value: f32) -> f32 {
+        match self.scale {
+            SliderScale::Linear => (value - self.min) / (self.max - self.min),
+            SliderScale::Logarithmic => {
+                let base = self.max / self.min;
+                (value / self.min).log(base).clamp(0.0, 1.0)
+            }
+        }
+    }
```
huacnlee: Move this 2 method to SliderScale.

crates/ui/src/slider.rs
```diff
@@ -168,6 +187,23 @@ impl SliderState {
         self
     }
 
+    /// Set the scale of the slider, default: Linear
+    pub fn scale(mut self, scale: SliderScale) -> Self {
```
huacnlee, line 236: Please also add this option to `SliderStory` for example.

crates/ui/src/slider.rs
```diff
@@ -123,6 +123,11 @@ impl SliderValue {
     }
 }
 
+pub enum SliderScale {
+    Linear,
+    Logarithmic,
+}
```
huacnlee, line 158: The was default value is `Logarithmic` right?

Here need impl `Default`, and `SliderState` new should be `SliderScale::default()`.

Moulberry: The previous behaviour was `Linear`, so I think that should be the default value

crates/ui/src/slider.rs
```diff
@@ -145,18 +151,31 @@ impl SliderState {
...
+            );
+        }
         self.min = min;
         self.update_thumb_pos();
         self
     }
 
     /// Set the maximum value of the slider, default: 100.0
     pub fn max(mut self, max: f32) -> Self {
+        if matches!(self.scale, SliderScale::Logarithmic) {
```
Moulberry: This check is good, but we can't rely on it.
Someone could do `.max(100.0).min(200.0)` and the check would be bypassed because of the order.

crates/ui/src/slider.rs
```diff
@@ -192,16 +228,42 @@ impl SliderState {
...
+                let base = self.max / self.min;
+                (base.powf(percentage) * self.min).clamp(self.min, self.max)
+            }
+        }
+    }
+
+    /// converts a value between the minimum and maximum value to a value between 0.0 and 1.0, depending on the chosen scale
+    fn value_to_percentage(&self, value: f32) -> f32 {
+        match self.scale {
+            SliderScale::Linear => (value - self.min) / (self.max - self.min),
```
Moulberry: This new code will produce NaNs when `self.max == self.min`

crates/ui/src/slider.rs
CherryWorm: Added the default and various other derives to SliderScale

crates/ui/src/slider.rs
CherryWorm: Forgot about that, fixed it

crates/ui/src/slider.rs
CherryWorm: I think the previous implementation just straight up didn't properly work for self.min != 0, which is why it didn't have this issue (except for when self.max == 0). What behavior would you prefer here? An invariant check when self.min/self.max are set, similar to how I implemented it for the log scale?

Moulberry: The previous behaviour was that if `max == min` (or `max == 0`, as you point out), the percentage would be zero.

crates/ui/src/slider.rs
CherryWorm: Added an example to the story where a logarithmic slider would be the right choice

https://github.com/user-attachments/assets/33d7cfb5-bb4a-43a9-a11d-c13cf8fe5de7

crates/ui/src/slider.rs
CherryWorm: https://github.com/longbridge/gpui-component/blob/b899100d09afa2d3a072b8c7b7364e5e7620f4b5/crates/ui/src/slider.rs#L207
As far as I can tell, the previous implementation had the exact same issue when self.max==0, it just didn't subtract self.min before dividing. Added a check to return 0 when the range is zero or negative now.

CherryWorm:
Actually I just realized that this inadvertently contains a breaking change.

Right now, `min` doesn't actually change the mapping between the value and the slider percentage, it only constrains the slider percentage to stay above `min/max`. I didn't realize this because this isn't really documented anywhere, and the api is fairly confusing here, as the name `min` suggests that it is the counterpart to `max`.

I propose moving `min` and `max` together with `percentage_to_value` and `value_to_percentage` to `SliderScale,` and instead having optional properties named `clamp_min` and `clamp_max` on the `SliderState`, that constrain the movement of the slider. That way people that may rely on the unintuitive behavior don't get their code silently broken.

huacnlee:
Hi @CherryWorm please also add this change to docs.

https://github.com/longbridge/gpui-component/blob/main/docs/docs/components/slider.md

crates/ui/src/slider.rs
```diff
@@ -132,6 +123,13 @@ impl SliderValue {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum SliderScale {
+    #[default]
+    Linear,
+    Logarithmic,
```
huacnlee, line 157: We need doc here to describe difference these 2 features.

crates/ui/src/slider.rs
```diff
@@ -201,16 +234,49 @@ impl SliderState {
...
 
+    /// converts a value between 0.0 and 1.0 to a value between the minimum and maximum value, depending on the chosen scale
+    fn percentage_to_value(&self, percentage: f32) -> f32 {
+        match self.scale {
+            SliderScale::Linear => self.min + (self.max - self.min) * percentage,
+            SliderScale::Logarithmic => {
+                // when percentage is 0, this simplifies to (max/min)^0 * min = 1 * min = min
+                // when percentage is 1, this simplifies to (max/min)^1 * min = (max*min)/min = max
+                // we clamp just to make sure we don't have issue with floating point precision
+                let base = self.max / self.min;
```
huacnlee, line 285: Here, if `min` is zero, will crash?

PR #1544, opened 2025-11-08
slider: Improve Slider interaction to click on bar to drag.

CherryWorm:
This pull request adds a drag event to the slider bar itself, analogous to the drag handles, in case the slider is not a range slider. This allows for a more intuitive interaction with the slider, where one can click and drag on any part of the slider bar container.

https://github.com/user-attachments/assets/4d41ba10-2992-4a42-beb6-0bc376185c5e

This doesn't change the behavior for range sliders, as it's ambiguous which slider should be dragged in case the event starts between the sliders. Though if desired, it wouldn't be difficult to implement the same behavior for clicks below the start or above the end slider.

### DIFF
--- a/crates/ui/src/slider.rs
+++ b/crates/ui/src/slider.rs
@@ -17,6 +17,15 @@ impl Render for DragThumb {
     }
 }
 
+#[derive(Clone)]
+struct DragSlider(EntityId);
+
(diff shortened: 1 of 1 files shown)


huacnlee:
Thanks.

PR #1548, opened 2025-11-09
Add ability to disable resizing of the window

Gbr22:
It is currently not possible to disable window resizing in gpui-component.

It would be nice if gpui-component inherited `is_resizable` from `WindowOptions`, but AFAIK, there is no way to access that value of the window after window creation, so I added the resizable option to `Root` (and `WindowBorder`) instead, with `true` as the default value to maintain the current behavior when unset.

If you are unhappy with the API in the PR, but agree that resizing should be something that can be disabled, then treat this as a feature request.

### DIFF
--- a/crates/ui/src/root.rs
+++ b/crates/ui/src/root.rs
@@ -219,6 +219,7 @@ pub struct Root {
     pub notification: Entity<NotificationList>,
     sheet_size: Option<DefiniteLength>,
     view: AnyView,
+    is_resizable: bool,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


huacnlee (changes requested):
We shouldn't add this; it seems we can simply check whether the current window allows resizing within the WindowBorder.

huacnlee:
I checked the GPUI window, and it doesn't seem to have anything related to it. 

However, your current modification still doesn't work; it only addresses the Linux limitation.

I think this requirement should be addressed by adding this parameter to the GPUI side.

PR #1551, opened 2025-11-10
select: Add delegate() getter

Moulberry:
Allow getting the SelectDelegate for a SelectState.

Unfortunately delegate_mut() isn't possible due to the GpuiBorrow guard.

### DIFF
--- a/crates/ui/src/select.rs
+++ b/crates/ui/src/select.rs
@@ -647,6 +647,11 @@ where
             .map(|item| item.value().clone());
     }
 
+    pub fn delegate<'a>(&self, cx: &'a App) -> &'a D {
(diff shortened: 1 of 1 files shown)


crates/ui/src/select.rs
```diff
@@ -647,6 +647,11 @@ where
             .map(|item| item.value().clone());
     }
 
+    pub fn delegate<'a>(&self, cx: &'a App) -> &'a D {
```
huacnlee, line 650: This API not match with the `List::delegate` and `delegate_mut` is also not.

So I don't think we can just add this one is good.

Moulberry:
I still think it is valuable to have a way to access the StateDelegate, but I understand

PR #1554, opened 2025-11-11
chore: Update GPUI and update `closest_index_for_x` API changes.

huacnlee:


huacnlee:
Close by https://github.com/zed-industries/zed/pull/42505 reverted.

PR #1560, opened 2025-11-11
resizable: Improve to avoid panels from improperly resizing on container resizing.

CherryWorm:
This pull request changes the internal behavior of the resizable panel to get rid of some weird behavior with many panels and resizing.

Previously, when having multiple panels and dragging them into each other, panels that shouldn't be affected got improperly resized:

https://github.com/user-attachments/assets/20d280c0-8a51-4e96-9c4a-3ade3deaca02

This pull request completely fixes this behavior:

https://github.com/user-attachments/assets/7c207dd8-d0a0-47d3-8d9e-c7127f43fa73

In addition, dragging the container smaller and larger again wasn't idempotent:

https://github.com/user-attachments/assets/ca54b039-fd79-4fa8-87cf-590c06dd2c18

This is also fixed by this pull request, replacing the behavior with more intuitive behavior, keeping the fractional size of all panels constant:

https://github.com/user-attachments/assets/365d0a99-fd4b-4948-bc1b-bafdd826394a

https://github.com/user-attachments/assets/b208b34f-18e7-42a4-825a-9dad421b7ad3

If desired, it would also be easy to only change the size of the panels that don't have a default size set when the container changes size.

### DIFF
--- a/crates/ui/src/resizable/panel.rs
+++ b/crates/ui/src/resizable/panel.rs
@@ -6,7 +6,8 @@ use std::{
 use gpui::{
     canvas, div, prelude::FluentBuilder, Along, AnyElement, App, AppContext, Axis, Bounds, Context,
     Element, ElementId, Empty, Entity, EventEmitter, InteractiveElement as _, IntoElement,
-    MouseMoveEvent, MouseUpEvent, ParentElement, Pixels, Render, RenderOnce, Style, Styled, Window,
+    IsZero as _, MouseMoveEvent, MouseUpEvent, ParentElement, Pixels, Render, RenderOnce, Style,
+    Styled, Window,
 };
 
 use crate::{h_flex, resizable::PANEL_MIN_SIZE, v_flex, AxisExt};
@@ -271,9 +272,13 @@ impl RenderOnce for ResizablePanel {
             // 3. initial_size is Some and size is Some, use `size`.
             .when(self.initial_size.is_none(), |this| this.flex_shrink())
             .when_some(self.initial_size, |this, initial_size| {
-                // We always set the panel to flex_none, to prevent unwanted movement
-                this.flex_none()
-                    .flex_basis(initial_size.min(size_range.end).max(size_range.start))
+                // The `self.size` is None, that mean the initial size for the panel,
+                // so we need set `flex_shrink_0` To let it keep the initial size.
+                this.when(
+                    panel_state.size.is_none() && !initial_size.is_zero(),
+                    |this| this.flex_none(),
+                )
+                .flex_basis(initial_size)
             })
             .map(|this| match panel_state.size {
                 Some(size) => this.flex_basis(size.min(size_range.end).max(size_range.start)),


huacnlee:
There may some incorrect change you need to check again. When we resize the window, the panels size scale will affect.


```bash
cargo run --example dock
```

https://github.com/user-attachments/assets/0e06b281-48ef-4de3-ab4d-e9e503202de7

huacnlee:
And in this case, there have new issue appeared (I have checked main branch, it was not exists.):

```bash
cargo run -- resizable
```

https://github.com/user-attachments/assets/eba32a23-b2ec-41bb-87e0-eb00715f887f

crates/ui/src/resizable/mod.rs
```diff
@@ -149,7 +153,7 @@ impl ResizableState {
 
     fn sync_real_panel_sizes(&mut self, _: &App) {
         for (i, panel) in self.panels.iter().enumerate() {
-            self.sizes[i] = panel.bounds.size.along(self.axis).floor();
+            self.sizes[i] = panel.bounds.size.along(self.axis);
```
huacnlee, line 157: I guess, this is the source issue of the panel have a bit size change on resizing, so maybe you can split this part as a single PR. Then we can merge it first.

han1548772930: When the overall window size changes, the size inside also changes.

https://github.com/user-attachments/assets/aa85f92b-737d-421e-9514-2ef83631f918

CherryWorm: > I guess, this is the source issue of the panel have a bit size change on resizing, so maybe you can split this part as a single PR. Then we can merge it first.

There wasn't just one issue, there were multiple. This one didn't affect anything too much, just caused a bit of jittering, and I decided to fix it if I'm touching the code anyways. I can split it out, but the value of this change on its own is pretty much 0. In fact the panels still have a slight size change on resizing, because the flex layouting engine sucks. If I had more experience with gpuis elements, I would've tried to rewrite the whole thing without flex.

CherryWorm:
> There may some incorrect change you need to check again. When we resize the window, the panels size scale will affect.
> 
> ```shell
> cargo run --example dock
> ```
> 
> Screen.Recording.2025-11-12.at.09.50.18.mov

This example doesn't seem to use the code path I modified at all. The dock uses the resizable handle, but not the resizable panels. Can you expand on what the issue here is?

CherryWorm:
> And in this case, there have new issue appeared (I have checked main branch, it was not exists.):
> 
> ```shell
> cargo run -- resizable
> ```
> 
> Screen.Recording.2025-11-12.at.09.55.13.mov

I unfortunately can't reproduce this issue at all. I actually had a very similar bug in my application on the old version that got fixed with the changes I made. I imagine this is just another weird rounding bug in the flex layouting engine (sometimes there's an off-by-one between flex items and there's either a 1px gap or they overlap by 1px)

https://github.com/user-attachments/assets/18bb25a2-1de0-475f-938a-d3726518494b

huacnlee:
> > There may some incorrect change you need to check again. When we resize the window, the panels size scale will affect.
> > ```shell
> > cargo run --example dock
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Screen.Recording.2025-11-12.at.09.50.18.mov
> 
> This example doesn't seem to use the code path I modified at all. The dock uses the resizable handle, but not the resizable panels. Can you expand on what the issue here is?

No, it used `ResizablePanelGroup`

https://github.com/longbridge/gpui-component/blob/8c2d4de4352e7eceef3b313d33948da79db16880/crates/ui/src/dock/stack_panel.rs#L421-L428

The problem caused by your changes is that if the window resizes, the proportions of the panels inside will be affected, and the panel sizes will gradually become equal. 

The correct approach is to always maintain the original proportions.

Try the `dock` example, and see the different of your version and main branch.

CherryWorm:
> > > There may some incorrect change you need to check again. When we resize the window, the panels size scale will affect.
> > > ```shell
> > > cargo run --example dock
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Screen.Recording.2025-11-12.at.09.50.18.mov
> > 
> > 
> > This example doesn't seem to use the code path I modified at all. The dock uses the resizable handle, but not the resizable panels. Can you expand on what the issue here is?
> 
> No, it used `ResizablePanelGroup`
> 
> https://github.com/longbridge/gpui-component/blob/8c2d4de4352e7eceef3b313d33948da79db16880/crates/ui/src/dock/stack_panel.rs#L421-L428
> 
> The problem caused by your changes is that if the window resizes, the proportions of the panels inside will be affected, and the panel sizes will gradually become equal.
> 
> The correct approach is to always maintain the original proportions.
> 
> Try the `dock` example, and see the different of your version and main branch.

Ah now I understand, I didn't realize you had two panels next to each other. This is actually intended behavior: when you reduce the window size beyond the minimum size of the two panels, they will stop shrinking. When you then expand the window again, they'll obviously have the same aspect ratio, as both have the same minimum size. If you don't reduce them to their minimum size, they'll always keep their aspect ratio, but when one stops shrinking while the window is made smaller, the aspect ratio obviously changes.

I first implemented this without the minimum size, but I assumed it would be desirable that panels don't shrink more than they're supposed to. This is just a matter of removing the clamping in https://github.com/longbridge/gpui-component/blob/b0de79ea23caf91e481385d7e8fe118ad0da3f7d/crates/ui/src/resizable/mod.rs#L237-L238. They still won't be physically smaller as the size is still clamped in the render function, but they will take on values outside the allowed range in the sizes vec. Up to you if that's desired.

huacnlee:
Look at Zed. Notice the panel on the right. Even after repeatedly scaling back to the original window size, the panel on the right still maintains its previous width ratio.

https://github.com/user-attachments/assets/1e634094-e469-497d-822a-ba9bff042fdf

huacnlee:
However, I just discovered that the GPUI Component `main` branch itself may have a problem. After multiple scaling operations, the scale will change, which may be a minor precision calculation issue.

Sorry, I might have misread it when I checked this morning; it seems you fixed the problem.

CherryWorm:
Yeah that was the second example I gave: 

https://github.com/user-attachments/assets/ca54b039-fd79-4fa8-87cf-590c06dd2c18

But I'm happy to remove the clamping, that should yield the behavior you want:

https://github.com/user-attachments/assets/9edc9553-58c8-439b-a654-3b445e9a39e4

huacnlee:
I've reviewed it several times.

It seems the new changes still haven't resolved the issue of width changes after resizing, affecting both the main branch and the version you modified.

Given this, the PR changes require further improvement and testing.

---

Panel resizing involves many details, and we did a lot of interactive test during our previous implementation.

Currently, there are indeed some known issues, which I really want we can be fixed that, but more testing and verification are needed; otherwise, it could easily introduce new problems.

CherryWorm:
Removed the clamping, the dock example now has the behavior you asked for and everything else still behaves as before. I'll maintain this branch for the foreseeable future, because resizables with more than 3 panels are clearly just straight up broken in multiple ways without these fixes. Happy to fix any issues you might have with it, but I need to know what the issues are in order to do that. I can't reproduce any aspect ratio changes on my end even when vigorously resizing the window for dozens of seconds.

huacnlee:
> And in this case, there have new issue appeared (I have checked main branch, it was not exists.):
> 
> ```shell
> cargo run -- resizable
> ```
> 
>  Screen.Recording.2025-11-12.at.09.55.13.mov

Now, only have this issue.

image

crates/ui/src/resizable/panel.rs
```diff
@@ -260,16 +271,12 @@ impl RenderOnce for ResizablePanel {
...
-                // The `self.size` is None, that mean the initial size for the panel,
-                // so we need set `flex_shrink_0` To let it keep the initial size.
-                this.when(
-                    panel_state.size.is_none() && !initial_size.is_zero(),
-                    |this| this.flex_none(),
-                )
-                .flex_basis(initial_size)
+                // We always set the panel to flex_none, to prevent unwanted movement
+                this.flex_none()
+                    .flex_basis(initial_size.min(size_range.end).max(size_range.start))
```
huacnlee: The last issue caused by this change, I tried to revert this, then bug is gone.

crates/ui/src/resizable/panel.rs
```diff
@@ -157,7 +158,18 @@ impl RenderOnce for ResizablePanelGroup {
...
-                        move |bounds, _, cx| state.update(cx, |state, _| state.bounds = bounds)
+                        move |bounds, _, cx| {
+                            state.update(cx, |state, cx| {
+                                let size_changed = state.bounds.size.along(self.axis)
+                                    != bounds.size.along(self.axis);
+
+                                state.bounds = bounds;
+
+                                if size_changed {
+                                    state.adjust_to_container_size(cx);
```
huacnlee, line 169: And I have tried to remove this line to test.

It is looking like no difference, so is this code is necessary?

huacnlee: Oh, I see the difference. Let's keep this.

huacnlee:
Hi @CherryWorm please check my commits, if you are not any question, I am going to merge this PR.

CherryWorm:
Thanks, this looks good!

huacnlee:
Thank you, panel resizing is better now.

PR #1572, opened 2025-11-12
input: Avoid blocking vertical scroll events in single-line mode (#1562)

### LINKED ISSUE #1562: input: Avoid block scroll vertical event when is single line more.

chulingera2025:
Fixes #1562

  ## Problem
  Single-line input fields were blocking all scroll wheel events, preventing parent containers from handling vertical page scrolling.

  ## Solution
  - Check if input is in single-line mode and scroll is purely vertical
  - Allow vertical scroll events to propagate in single-line mode
  - Only stop propagation when scroll offset actually changes

  ## Behavior
  - Single-line inputs: vertical scroll passes through to parent
  - Multi-line inputs: all scroll behavior unchanged
  - Horizontal scrolling still works in both modes

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -6,10 +6,9 @@ use anyhow::Result;
 use gpui::{
     actions, div, point, prelude::FluentBuilder as _, px, Action, App, AppContext, Bounds,
     ClipboardItem, Context, Entity, EntityInputHandler, EventEmitter, FocusHandle, Focusable,
-    InteractiveElement as _, IntoElement, IsZero, KeyBinding, KeyDownEvent, MouseButton,
-    MouseDownEvent, MouseMoveEvent, MouseUpEvent, ParentElement as _, Pixels, Point, Render,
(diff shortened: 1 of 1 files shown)


crates/ui/src/input/state.rs
```diff
@@ -1291,15 +1292,27 @@ impl InputState {
...
             .map(|layout| layout.line_height)
             .unwrap_or(window.line_height());
         let delta = event.delta.pixel_delta(line_height);
-        self.update_scroll_offset(Some(self.scroll_handle.offset() + delta), cx);
+
+        // In single-line mode, if this is a pure vertical scroll, don't block
+        // the event so parent containers can handle page scrolling
+        if self.mode.is_single_line() && delta.x.is_zero() && !delta.y.is_zero() {
+            return;
+        }
```
huacnlee: I think we not need this part. Have you tried to test is the below `if self.scroll_handle.offset() != old_offset` will handle all?

chulingera2025: Indeed, when y = 0 in single-line mode, there are no conditional branches, and the simplified logic correctly handles all cases.

huacnlee:
Thank you.

PR #1578, opened 2025-11-12
notification: Fix close animation not working when closed from content

Moulberry:
Currently, if dismiss is called from the content_builder or action_builder, the closing animation doesn't play.

This is because the callback happens after the `closing` variable is set, so the animation will only start on the next frame, (and for some reason notify() doesn't trigger another frame?).

This change makes it so that the content_builder is called first, so the closing animation starts playing immediately on the same frame.

This also adds a check to dismiss to prevent spawning unnecessary futures if it is called multiple times

### DIFF
--- a/crates/ui/src/notification.rs
+++ b/crates/ui/src/notification.rs
@@ -230,6 +230,9 @@ impl Notification {
 
     /// Dismiss the notification.
     pub fn dismiss(&mut self, _: &mut Window, cx: &mut Context<Self>) {
+        if self.closing {
+            return;
+        }
         self.closing = true;
         cx.notify();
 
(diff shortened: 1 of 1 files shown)


huacnlee:
Please submit the video, I just tested example, the red arrow point button have `action`, there have a "Retry" button will call dismiss from internal, the close animation is look ok.

image

https://github.com/user-attachments/assets/fcfaaca8-9c0c-4d06-9d57-3669b8de851f

Moulberry:
> Please submit the video, I just tested example, the red arrow point button have `action`, there have a "Retry" button will call dismiss from internal, the close animation is look ok.

The retry button uses cx.listener, it doesn't call Notification#dismiss directly from the content_builder callback.

e.g. I have code that looks like this:

```rust
let notification = Notification::new()
        .autohide(false)
        .content(move |notification, window, cx| {
            if modal_action.get_finished_at().is_some() {
                notification.dismiss(window, cx);
            }
```

The dismiss is called directly from inside the content_builder callback.

Here is a video:

https://youtu.be/tYjki26OmRY

PR #1580, opened 2025-11-12
dock: Add a way for Panels to track the tab bar they belong to

CherryWorm:
Currently, there is no good way for a Panel in a dock to e.g. add a tab to its own tab bar. This is especially annoying as the state in DockArea::items isn't synced to the underlying views, meaning the `DockArea::add_panel` function will do nothing after first splitting the panel, and then deleting the only element in the first part of the split.

https://github.com/user-attachments/assets/2c61f2f0-9b1c-4154-8a53-b917335b94f0

This pull request adds a function that is called when a PanelView is added to a tabbar via `add_panel` or `insert_panel_at` or when it is removed via `detach_panel`. 

This allows for entities that implement Panel to keep track of and directly access the tab bar they belong to, and for example add new panels to it through buttons in the dropdown menu or the top right of the tab bar.

### DIFF
--- a/crates/ui/src/dock/panel.rs
+++ b/crates/ui/src/dock/panel.rs
@@ -118,16 +118,11 @@ pub trait Panel: EventEmitter<PanelEvent> + Render + Focusable {
     /// Only current Panel will touch this method.
     fn set_zoomed(&mut self, zoomed: bool, window: &mut Window, cx: &mut App) {}
 
-    /// Set the tab panel this panel belongs to
-    ///
-    /// This method will be called when the panel is added to or removed from a tab panel.
-    fn set_tab_panel(
-        &mut self,
-        tab_panel: Option<WeakEntity<TabPanel>>,
-        window: &mut Window,
-        cx: &mut App,
-    ) {
-    }
+    /// When this Panel is added to a TabPanel, this will be called.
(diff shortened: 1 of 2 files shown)


CherryWorm:
For reference, this is what I'm intending to use this change for

https://github.com/user-attachments/assets/1dae0efa-56c2-4032-9d50-7409bd58b6cc

huacnlee (changes requested):
The delegate method `set_tab_panel` is a bit confuse.

~~I think this should be use `PanelEvent` to do.~~

~~https://github.com/longbridge/gpui-component/blob/57251a5ac022bd7a363d1606810869c121df6197/crates/ui/src/dock/panel.rs#L13-L17~~

```rs
AddedTo {
  parent: Arc<dyn PanelView>
},
DetachedFrom {
  parent: Arc<dyn PanelView>
},
```

---

Or, if we split as `on_removed`, `on_added_to` it better.

```rs
/// When this Panel is added to a TabPanel, this will be called.
fn on_added_to(&mut self, tab_panel: Entity<TabPanel>, window: &mut Window, cx: &mut App);
/// When this Panel is removed from a TabPanel, this will be called.
fn on_removed(&mut self, window: &mut Window, cx: &mut App);
```

CherryWorm:
Split the api as requested. I used a WeakEntity instead though, as back references should always be weak

PR #1588, opened 2025-11-13
resizable, dock: Improve Panel adjust size logic on layout changed.

CherryWorm:
This is a follow up to #1560 I noticed a couple small issues with resizable panels and their use in docks after experimenting a lot:
- Sometimes the resizable group would take about ~500ms to update after a resize. This was caused by the `StackPanel` not getting notified and therefore not redrawn when the `ResizableState` was notified, as well as the notify in `adjust_to_container_size` not actually firing properly. The latter was fixed by defering the notify, though I'm still not quite sure why that is necessary.
- In a couple of places, the total size all the panel missmatches the container size, causing small glitches the first time a panel is resized. Update the code in those places to properly adjust all the panel sizes so they match the container size.

### DIFF
--- a/crates/ui/src/dock/stack_panel.rs
+++ b/crates/ui/src/dock/stack_panel.rs
@@ -61,10 +61,6 @@ impl StackPanel {
             cx.subscribe(&state, |_, _, _: &ResizablePanelEvent, cx| {
                 cx.emit(PanelEvent::LayoutChanged)
             }),
-            // Bubble up changes in the resizable state to force a rerender.
-            cx.observe(&state, move |_, _, cx| {
-                cx.notify();
-            }),
         ];
 
         Self {


CherryWorm:
I verified that the resizable story behaves identical to main, and docks now behave a lot better. Let me know if there's any issue or regression I didn't find

huacnlee:
Hi @CherryWorm thanks for improve.

But I have done a little change. 

And I removed `defer_in` and `observe` to notify. These two parts seem unnecessary and may be caused by some other issues, but these two implementations are not quite right.

I have a test after remove that 2 parts, the test show is all thing good.

CherryWorm:
Thanks for the changes! Part of that change is definitely necessary though, here's a video of the issue I'm talking about:

https://github.com/user-attachments/assets/26712d1d-d60f-40a6-8aca-0eb81250ab6b

The StackPanel consistently takes about ~500ms until it's redrawn again after it has been resized, leading to layout jumps.

I did some more experimenting and it turns out bubbling up the notify isn't necessary, but defering the notify is. Without it, this issue still happens. 

An overall better fix that would also improve responsiveness in general would be to not use canvas to update the bounds, but instead wrap the whole thing in a custom Element, but that would require more extensive refactoring.

huacnlee:
I have just done a test with this commit: https://github.com/longbridge/gpui-component/pull/1588/commits/24d8ef6bbc2a4a0a77daa5615d51bc95b5b6a65c 

The issue of your video show also exists.

https://github.com/user-attachments/assets/9a4b4079-06d9-4bdc-83e8-e34c3274541d

CherryWorm:
I just spent the last 15 minutes dragging around different windows in the dock example and continuously closing and opening the side dock again, and I can't reproduce this :/ It works exactly as it should every single time. I had the same experience in the application I'm currently working on.

https://github.com/user-attachments/assets/dd3f2e95-e150-4f70-bbd2-e046d61ef072

huacnlee:
Ok, let's ignore it.

CherryWorm:
But this latest version definitely has the issue. 

Have you tried in release mode on your end? What can be seen in your video might just be the expected delay that comes from updating bounds using the canvas, which always introduces a small delay (that might be big in dev mode with a lot of stuff on screen on a not-so-fast computer). In my testing the resize sometimes took almost a second until it occurred without the defer, even in release mode.

huacnlee:
I have just test on `main` branch with `--release`. It looks like that issue not appear in release mode.

CherryWorm:
It definitely still happens on main, even in release mode

https://github.com/user-attachments/assets/4f6a4335-3050-42b6-ac6f-559d12e7f9c1

It also makes complete sense why this happens, the resizing happens in the canvases prepaint function, where I'm guessing notifying other entities just doesn't do anything in general.

CherryWorm:
Would you accept a pull request that defers the notify in the canvas callback? That way it only happens at the place it really needs to

PR #1591, opened 2025-11-13
input: Fix x offset resetting for single line inputs

### LINKED ISSUE #1589: input: `update_scroll_offset` resets the offset for single line inputs
When you have a single line input the scroll resets to the beginning when the text overflows (grows), and you cannot navigate to the end.
Seems like it was introduced here: https://github.com/longbridge/gpui-component/commit/6dd857a7f707f1539ac9652fc958869a5820f87c

This is from the popover story,

https://github.com/user-attachments/assets/0baf1937-2229-40c5-ae45-570fd758317c

zanmato:
Fixes #1589

Before:
https://github.com/user-attachments/assets/ae1dd1c4-de97-4336-94ab-e545a1e97047

After:
https://github.com/user-attachments/assets/62b5cf08-0782-426e-b2ee-dc018f140149

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -952,7 +952,9 @@ impl Element for TextElement {
         );
 
         let mut longest_line_width = wrap_width.unwrap_or(px(0.));
-        if state.mode.is_single_line() || !state.soft_wrap || lines.len() > 1 {
+        // 1. Single line
+        // 2. Multi-line with soft wrap disabled.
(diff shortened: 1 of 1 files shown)


zanmato:
This caused a regression with shift scrolling

zanmato:
The issue was that `scroll_size` was always `0` for single line inputs and no wrap inputs, causing the scroll to reset.

huacnlee:
Thank you for addresses this.

PR #1611, opened 2025-11-15
slider: Notify on bound update

CherryWorm:
The slider has the same issue that was fixed in #1588. When resizing, the state isn't notified when the bounds change, leading to the slider not updating until it is rendered again:

https://github.com/user-attachments/assets/de79c338-d489-4173-a0be-45af30c8ae95

Again, I believe the defer is necessary here. Only notifying without deferring does not change this behavior. With the deferred notify, this behavior is fixed:

https://github.com/user-attachments/assets/6e77dc52-ec1f-47a8-945f-4a70ce7bd48d

### DIFF
--- a/crates/ui/src/slider.rs
+++ b/crates/ui/src/slider.rs
@@ -656,7 +656,17 @@ impl RenderOnce for Slider {
                             .child({
                                 let state = self.state.clone();
                                 canvas(
-                                    move |bounds, _, cx| state.update(cx, |r, _| r.bounds = bounds),
+                                    move |bounds, _, cx| {
+                                        state.update(cx, |r, cx| {
+                                            r.bounds = bounds;
+                                            let entity = cx.entity();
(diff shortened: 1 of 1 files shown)


huacnlee:
I understand that this is indeed a problem, but using defer to fix it is not the right approach.

Have you tried `cargo run -- slider` in GPUI Component story, I'm seeing is that there is a slight delay after resizing, but not as slow as when you resize the video.

zanmato:
@CherryWorm Which distro/compositor are you using? I'm on Ubuntu 25.04/Wayland and I don't see the delay either. There is another issue which mentions laggy UI, and they're using XFCE/X11. But I can't replicate that either

huacnlee:
Not have solution for this, I am close this PR. 

And I'll see if I can resolve this issue when I have time to reopen a new PR.

PR #1631, opened 2025-11-18
Add Thai translations for UI elements

siteslave:


madcodelife:
I'm not sure if this is the best way to solve translations configuration. This is not a direction that can be sustainably maintained. 

In the future, we should consider a more flexible approach, such as using a dictionary for overrides.

PR #1640, opened 2025-11-18
icon: Rework icons to make use of the `IconNamed` trait

### LINKED ISSUE #1627: Icon rework
I have been looking over the approach to icons for a bit and I think there are some improvements to be made, this issue contains my suggestions.

The current implementation is pretty "locked" in its implementation, as it maps 82 enum variants to embedded asset paths. This leaves two main issues:
1. If you want more than 82 icons in your application, you have a problem
2. If you want an icon in your application which does not have an appropriate `IconName`, you must pick a "spare" option and remember this mapping throughout your application

Another, but somewhat separate issue is that multiple components utilize icons internally, which are only rendered when the icon is present in the loaded assets, even though the button that "contains" the icon is still rendered. My best example of this is the close button for a sheet, which is just a `Ghost` variant button with no text or icon, making it difficult, if not very unlikely that you actually discover (unless you subscribe to log events and see the errors for failing to get the icon). I personally think it would make sense to always have these icons available in the crate, as they are "essential" in some way.

My suggestion is to add a new public trait that requires a `.path` method to be implemented. The `Icon` is changed to take anything that implements this trait as input. Then this trait can be utilized by users of this crate to implement their "own" `IconName`, which can contain however many icons they like, with whatever name they desire. The current `IconName` from this crate can still be kept as a sensible default.

This way, the essential icons could also be embedded and enabled elsewhere, outside of the current `IconName`, to ensure they would not interfere with users of this crate (although that may go against the desired behaviour of overwriting these essential icons).

I think an implementation of just the trait would be quite simple, and if it sounds reasonable and is desired I can implement it.

andreaskrath:
This makes use of the `IconNamed` trait and a blanked implementation to convert anything that implements this to an `Icon`.

This allows for easily defined custom versions of `IconName`, while minimally changing existing code (essentially only if you previously made use of the `.path()` method on the `IconName` enum; this now requires an import of the `IconNamed` trait).

# Example
```rust
use gpui_component::IconNamed;

pub enum IconName {
    Encounters,
    Monsters,
    Spells,
}

impl IconNamed for IconName {
    fn path(self) -> gpui::SharedString {
        match self {
            IconName::Encounters => "icons/encounters.svg",
            IconName::Monsters => "icons/monsters.svg",
            IconName::Spells => "icons/spells.svg",
        }
        .into()
    }
}

// this allows for the following interactions (works with anything that has the `.icon(icon)` method
Button::new("my-button").icon(IconName::Spells);
Icon::new(IconName::Monsters); 
```

If you want to directly "render" a custom `IconName` you must implement the `RenderOnce` trait and derive `IntoElement` on the `IconName`.

```rust
use gpui::{IntoElement, RenderOnce};
use gpui_component::IconNamed;

#[derive(IntoElement)]
pub enum IconName {
    // The same as before
}

impl IconNamed for IconName {
    // The same as before
}

impl RenderOnce for IconName {
    fn render(self, _: &mut gpui::Window, _: &mut gpui::App) -> impl gpui::IntoElement {
        gpui_component::Icon::empty().path(self.path())
    }
}

// this allows for the following interaction
div()
    .child(IconName::Monsters)
```

Overall I think is an improvement to the existing way to do custom `IconName` implementations.

I am unsure if this change should also be reflected in the documentation on the section with "Icons & Assets", though I personally think it would make sense to highlight this way to do custom versions of `IconName` as it is considerably less involved than the current approach.

Closes #1627.

### DIFF
--- a/docs/docs/components/icon.md
+++ b/docs/docs/components/icon.md
@@ -7,6 +7,8 @@ description: Display SVG icons with various sizes, colors, and transformations.
 
 A flexible icon component that renders SVG icons from the built-in icon library. Icons are based on Lucide.dev and support customization of size, color, and rotation. The component requires SVG files to be provided by the user in the assets bundle.
 
+Before you start, please make sure you have read: [Icons & Assets](../assets.md) to understand how use SVG in GPUI & GPUI Component application.
+
 ## Import
 
 ```rust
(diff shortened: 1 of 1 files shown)


huacnlee:
Thank you. I have also updated the docs for this.

Issue #1489, opened 2025-11-03
Support lazy loading and `popup_menu` for Tree.


Trojanking123:
These two features are excellent for a mature tree component:
1. Lazy loading: Sub-items load only when you click or expand a folder.
2. Right-click context menu: Rename or perform other actions on any item.

just like this。
<img width="704" height="1065" alt="Image" src="https://github.com/user-attachments/assets/90d3fa86-63e3-4d89-9176-2339b66f7492" />

huacnlee:
Outdated issue, close for now.

PR #1644, opened 2025-11-19
tree: Add context menu support

### LINKED ISSUE #1643: docs(tree): ListItem doesn't have on_secondary_mouse_down
ListItem doesn't have `on_secondary_mouse_down`, docs: https://github.com/longbridge/gpui-component/blob/main/docs/docs/components/tree.md?plain=1#L154-L166

Should it be added to ListItem?
```rust
.when_some(self.on_secondary_mouse_down, |this, on_mouse_down| {
    this.on_mouse_down(MouseButton::Right, move |event, window, cx| {
        (on_mouse_down)(event, window, cx)
    })
})
```

zanmato:
Fixes #1643 and partially #1489

Adds `.context_menu` for `ListItem`

### DIFF
--- a/docs/docs/components/settings.md
+++ b/docs/docs/components/settings.md
@@ -454,5 +454,5 @@ Settings::new("app-settings")
 [SettingItem]: https://docs.rs/gpui-component/latest/gpui_component/setting/struct.SettingItem.html
 [SettingField]: https://docs.rs/gpui-component/latest/gpui_component/setting/enum.SettingField.html
 [NumberFieldOptions]: https://docs.rs/gpui-component/latest/gpui_component/setting/struct.NumberFieldOptions.html
-[GroupBox]: ./group_box.md
+[GroupBox]: ./group-box.md
 [Sizable]: https://docs.rs/gpui-component/latest/gpui_component/trait.Sizable.html


zanmato:
After making this I'm not sure if it's needed, or the docs could be updated to add context_menu on the list item child instead?

```rust
tree(&tree_state, |ix, entry, selected, window, cx| {
    ListItem::new(ix)
        .selected(selected)
        .child(
            div()
                .child(entry.item().label.clone())
                .context_menu(|menu, _, _| {
                    menu.menu("Rename", Box::new(Rename))
                        .menu("Delete", Box::new(Delete))
                        .separator()
                        .menu("Copy Path", Box::new(CopyPath))
                })
        )
})
```

zanmato:
There is a bug with this approach, you have to right click twice, because it seems like the `window.on_mouse_event` in ContextMenu fires before the on_mouse_down event of the table (or tree in this case). You can see it in the table_story. Probably because the uniform list is further down in the element tree than the context menu on the parent/root element?

zanmato:
It also means that it is showing the wrong context menu, since the right clicked index is the previous one :disappointed: Perhaps context menu needs to be something "global" like dialog/sheet? So we can force it open in the correct event listener?

zanmato:
@huacnlee Updated it to use the same pattern as table with a delegate. It suffers from the same issue as table though, see #1651 

Should we keep `tree()`?

huacnlee:
No, we can not change the `tree` API.

You're doing this effected this PR more complex now.

Small improve must be a small changes.

---


I don't know how to do this better right now, but I am sure we can't to change the `Tree` API. I will consider this when have a rest time.

docs/docs/components/settings.md
```diff
@@ -454,5 +454,5 @@ Settings::new("app-settings")
 [SettingItem]: https://docs.rs/gpui-component/latest/gpui_component/setting/struct.SettingItem.html
 [SettingField]: https://docs.rs/gpui-component/latest/gpui_component/setting/enum.SettingField.html
 [NumberFieldOptions]: https://docs.rs/gpui-component/latest/gpui_component/setting/struct.NumberFieldOptions.html
-[GroupBox]: ./group_box.md
+[GroupBox]: ./group-box.md
```
huacnlee, line 457: Split this into a single PR, so we can merge first.

zanmato:
> No, we can not change the `tree` API.
> 
> You're doing this effected this PR more complex now.
> 
> Small improve must be a small changes.
> 
> I don't know how to do this better right now, but I am sure we can't to change the `Tree` API. I will consider this when have a rest time.

OK! I will close this and open a PR with the docs fix. We could do `tree(...).context_menu({})` without breaking the API, but that would require storing the context menu handler in the `TreeState`.

PR #1651, opened 2025-11-20
menu: Use defer to avoid race conditions with click listeners

zanmato:
When using `.context_menu` on a parent element, the `window.on_mouse_event` would fire before, for example, the tables `on_mouse_down` so it'd be like this:
1. First right click = no context menu, right_clicked_row is set
2. Second right click = context menu, but with the row from the first click

Before: (Look at `Selected row`)  

https://github.com/user-attachments/assets/5eb89a00-f1ce-4423-b8d7-3ec61b848b83

After:

https://github.com/user-attachments/assets/10033ca4-4c2c-47b5-ace5-6ba745fcfee5


zanmato:
I also experimented with a global version in root with `render_context_menu_layer` and `window.open_context_menu_at`, which worked, but this change is less intrusive

huacnlee:
Thank you.

PR #1663, opened 2025-11-22
Add support for registering built-in themes and theme sets

dparnell:


huacnlee:
I don't think this is necessary, the built-in theme are supports to load theme from a folder by you want. Why do we need to add a ThemeSet?

dparnell:
this allows me to ship a theme built in without writing anything else to the disk

huacnlee:
I think, in this case, you can have your own ThemeRegistry.

The theme system of GPUI Component not requirement we must use `ThemeRegistry`, the `ThemeRegistry` is design to help us easy to watch a theme folder and auto reload. As you code and case that you described, this is not the target of the `ThemeRegistry` (For example when use used `register_theme_set` it will not have reload feature).

So, this change I don't want to introduce to.

dparnell:
ok, no problem.

PR #1706, opened 2025-11-30
input: Fix wrong font being used in Input.

### LINKED ISSUE #1665: Input: wrong wrap width when using a wide character font
When you specify a font family that has a wide character width for an `Input` element with `soft_wrap` being `true`, the wrap width is wrong and lines appear longer than the input width.

This only happens when specifying the font family directly in the input. When you specify the font family in its parent, it works correctly.

(The font used in the screenshot below is [IBM Plex Mono](https://fonts.google.com/specimen/IBM+Plex+Mono))

## **When specifying `font_family` directly in the `Input` element**

Image

## **When specifying `font_family` in the parent element**

Image

zanmato:
Closes #1665

## Description

When `Input::new().font_family("font")` is used, the font hasn't been applied yet in `Input::render`, therefore the `text_wrapper` will calculate using the wrong font. This is why it's working when the font is added to the parent element.

The fix moves the state update to `element` prepaint, where `window.text_style()` has the correct styles. The only downside is that there are now one additional state update.

## Screenshot

| Screenshot From 2025-11-30 20-03-11 | Screenshot From 2025-11-30 20-00-23 |

## How to Test

Use some wide font, like `IBM Plex Mono`. Apply it to `Input::new(&some_state).font_family("IBM Plex Mono")` where the input state has `soft_wrap`. Here is a diff for `textarea_story.rs`.

```diff
diff --git a/crates/story/src/textarea_story.rs b/crates/story/src/textarea_story.rs
index f2c63d44..bd5cbb31 100644
--- a/crates/story/src/textarea_story.rs
+++ b/crates/story/src/textarea_story.rs
@@ -77,9 +77,9 @@ impl TextareaStory {
 
         let textarea_no_wrap = cx.new(|cx| {
             InputState::new(window, cx)
-                .multi_line(true)
                 .rows(6)
-                .soft_wrap(false)
+                .multi_line(true)
+                .soft_wrap(true)
                 .default_value("This is a very long line of text to test if the horizontal scrolling function is working properly, and it should not wrap automatically but display a horizontal scrollbar.\nThe second line is also very long text, used to test the horizontal scrolling effect under multiple lines, and you can input more content to test.\nThe third line: Here you can input other long text content that requires horizontal scrolling.\n")
         });
 
@@ -195,7 +185,7 @@ impl Render for TextareaStory {
             .child(
                 section("Auto Grow")
                     .max_w_md()
-                    .child(Input::new(&self.textarea_auto_grow)),
+                    .child(Input::new(&self.textarea_auto_grow).font_family("IBM Plex Mono")),
             )
             .child(
                 section("Auto Grow with No Wrap")
```

## Checklist

### DIFF
--- a/crates/ui/src/styled.rs
+++ b/crates/ui/src/styled.rs
@@ -438,7 +438,7 @@ impl<T: Styled> StyleSized<T> for T {
             Size::XSmall => self.text_xs(),
             Size::Small => self.text_sm(),
             Size::Medium => self.text_sm(),
-            Size::Large => self.text_lg(),
+            Size::Large => self.text_base(),
             Size::Size(size) => self.text_size(size * 0.875),
         }
     }


crates/ui/src/input/state.rs
```diff
@@ -364,15 +365,12 @@ impl InputState {
 
         let text_style = window.text_style();
         let mouse_context_menu = MouseContextMenu::new(cx.entity(), window, cx);
+        let font_size = text_style.font_size.to_pixels(window.rem_size());
```
huacnlee: Here is not correct, I am sure.

The font_size should use a relative value to store in state.

We only can use `.to_pixels(window.rem_size());` in render stage.

zanmato: I've updated it to store Rems instead

PR #1736, opened 2025-12-03
editor: Add InlineCompletion LSP support

duanebester:
Closes [1729](https://github.com/longbridge/gpui-component/issues/1729)

## Description

Adds support for inline completions, which allows the editor to support agentic "tab to complete" solutions. This works, (tested macOS), but not sure of any gotchas or different approaches.

## Screenshot

https://github.com/user-attachments/assets/8e437f1e-9792-4417-b070-0a1dd8de6024

## How to Test

User implements the `inline_completion` method in CompletionsProvider.

Not implementing should have no effect. 

## Checklist

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -812,9 +812,10 @@ pub(super) struct PrepaintState {
     bounds: Bounds<Pixels>,
     // Inline completion rendering data
     /// Shaped ghost lines to paint after cursor row (completion lines 2+)
-    inline_completion_ghost_lines: Vec<ShapedLine>,
+    ghost_lines: Vec<ShapedLine>,
     /// First line of inline completion (painted after cursor on same line)
-    inline_completion_first_line: Option<ShapedLine>,
+    ghost_first_line: Option<ShapedLine>,
+    ghost_lines_height: Pixels,
 }
 
 impl PrepaintState {
@@ -1095,16 +1096,15 @@ impl Element for TextElement {
         }
         last_layout.lines = Rc::new(lines);
 
-        // Compute inline completion ghost lines before scroll_size calculation
-        let (inline_completion_first_line, inline_completion_ghost_lines) =
-            Self::layout_inline_completion(
-                state,
-                &last_layout.visible_range,
-                text_size,
-                window,
-                cx,
-            );
-        let ghost_line_count = inline_completion_ghost_lines.len();
+        let (ghost_first_line, ghost_lines) = Self::layout_inline_completion(
+            state,
+            &last_layout.visible_range,
+            text_size,
+            window,
+            cx,
+        );
+        let ghost_line_count = ghost_lines.len();
+        let ghost_lines_height = ghost_line_count as f32 * line_height;
 
         let total_wrapped_lines = state.text_wrapper.len();
         let empty_bottom_height = if state.mode.is_code_editor() {
@@ -1116,9 +1116,6 @@ impl Element for TextElement {
         } else {
             px(0.)
         };
-
-        // Add ghost line height to scroll size
-        let ghost_lines_height = ghost_line_count as f32 * line_height;
 
         let scroll_size = size(
             if longest_line_width + line_number_width + RIGHT_MARGIN > bounds.size.width {
@@ -1236,8 +1233,9 @@ impl Element for TextElement {
             hover_definition_hitbox,
             document_color_paths,
             indent_guides_path,
-            inline_completion_first_line,
-            inline_completion_ghost_lines,
+            ghost_first_line,
+            ghost_lines,
+            ghost_lines_height,
         }
     }
 
@@ -1364,7 +1362,7 @@ impl Element for TextElement {
 
         // Paint text with inline completion ghost line support
         let mut offset_y = mask_offset_y + invisible_top_padding;
-        let ghost_lines = &prepaint.inline_completion_ghost_lines;
+        let ghost_lines = &prepaint.ghost_lines;
         let has_ghost_lines = !ghost_lines.is_empty();
 
         for (ix, line) in prepaint.last_layout.lines.iter().enumerate() {
@@ -1413,16 +1411,13 @@ impl Element for TextElement {
         let mut offset_y = px(0.);
         if let Some(line_numbers) = prepaint.line_numbers.as_ref() {
             offset_y += invisible_top_padding;
-            let ghost_line_count = prepaint.inline_completion_ghost_lines.len();
 
-            // Paint line number background (extend for ghost lines)
-            let total_ghost_height = ghost_line_count as f32 * line_height;
             window.paint_quad(fill(
                 Bounds {
                     origin: input_bounds.origin,
                     size: size(
                         prepaint.last_layout.line_number_width - LINE_NUMBER_RIGHT_MARGIN,
-                        input_bounds.size.height + total_ghost_height,
+                        input_bounds.size.height + prepaint.ghost_lines_height,
                     ),
                 },
                 cx.theme().editor_background(),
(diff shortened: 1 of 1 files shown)


crates/story/examples/editor.rs
```diff
@@ -187,6 +187,35 @@ impl CompletionProvider for ExampleLspStore {
...
+    ) -> Task<Result<Option<SharedString>>> {
+        let text_string = text.to_string();
+
+        cx.background_spawn(async move {
+            // Get the current line text before cursor
+            let line_start = text_string[..offset]
+                .rfind('\n')
+                .map(|i| i + 1)
+                .unwrap_or(0);
+            let current_line = &text_string[line_start..offset];
```
duanebester: Is there a better way to do this?

crates/ui/src/input/lsp/completions.rs
```diff
@@ -145,4 +181,107 @@ impl InputState {
...
+                }
+
+                Some(provider.inline_completion(&text, offset, window, cx))
+            })?;
+
+            let Some(task) = task else {
+                return Ok(());
+            };
+
+            let response = task.await?;
```
duanebester, line 254: Is there a better way to kick off and wait on a task?

crates/story/examples/editor.rs
huacnlee: This is ok, at least in story.

huacnlee: There have a RopeExt for get a line.

crates/ui/src/input/lsp/completions.rs
```diff
@@ -27,6 +30,35 @@ pub trait CompletionProvider {
...
+        _window: &mut Window,
+        _cx: &mut Context<InputState>,
+    ) -> Task<Result<Option<SharedString>>> {
+        Task::ready(Ok(None))
+    }
+
+    /// Returns the debounce duration for inline completions.
+    ///
+    /// Default: 300ms
+    fn inline_completion_debounce(&self) -> Duration {
```
huacnlee, line 66: Add `#[inline]` here

crates/ui/src/input/state.rs
```diff
@@ -331,6 +331,10 @@ pub struct InputState {
     _subscriptions: Vec<Subscription>,
 
     pub(super) _context_menu_task: Task<Result<()>>,
+    /// Text to display as an inline completion suggestion (ghost text)
+    pub(super) inline_completion_text: Option<SharedString>,
+    /// Task for debouncing inline completion requests
+    pub(super) inline_completion_task: Task<Result<()>>,
```
huacnlee: Move this part, into a sub struct named `struct InlineCompletion` in `input/completion.rs`.

```rs
struct InlineCompletion {
  text: SharedString,
  _task: Task<Result<()>>
}
```

crates/ui/src/input/lsp/completions.rs
```diff
@@ -27,6 +30,35 @@ pub trait CompletionProvider {
...
+    /// The provider can analyze the text and cursor position to determine
+    /// what inline completion suggestion to show.
+    ///
+    /// Returns `Ok(Some(text))` to show a suggestion, or `Ok(None)` for no suggestion.
+    /// Default implementation returns `None`.
+    ///
+    /// # Arguments
+    /// * `text` - The current text content
+    /// * `offset` - The cursor position in bytes
+    fn inline_completion(
```
huacnlee, line 51: Add `LSP` of `textDocument/inlineCompletion` doc link here.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_inlineCompletion

crates/ui/src/input/state.rs
duanebester: Would `input/completion.rs` be a new file?

huacnlee: crates/ui/src/input/lsp/completions.rs

crates/ui/src/input/lsp/completions.rs
duanebester: Should I convert this to be: `Task<Result<CompletionResponse>>` as well to match the spec?

crates/ui/src/input/state.rs
duanebester: If the `InlineCompletion` struct could be swapped/destroyed easily, the task will be dropped and this is all okay, correct?

huacnlee: You can do as you think, I just say this can move to an external file to have a tidy code.

crates/ui/src/input/lsp/completions.rs
duanebester: Ah, `InlineCompletionResponse` requires
```
lsp-types = { version = "0.97.0", features = ["proposed"] }
```

huacnlee:
image

image


The inline completion text has not render correct.

huacnlee:
In Zed, there have a DiaplayMap for render, that is difference like the actual text. But it is a bit complex, I have not understood it all.

In this case, I think we must insert the inline completion text to the `text`. Or we have a holder (Maybe we can do like the password mask (I am not sure if it is possible).

### For example:

If we have added an empty (Only keep `\s`, and with same length of the inline completion text) placeholder into the source text before `shape_line`. After this, the primary text paint will leave a space for us to paint the inline completion. 

But this you must handle more details, the `shape_line` will keep the `lines` in `LastLayout` for selection, position cursor on mouse click ..., we need to ensure if user accept or not of the inline completion, the `LastLayout` will have a correct info.

https://github.com/longbridge/gpui-component/blob/26ea1576bf744fbb0e5a6bdc234921121cd5722d/crates/ui/src/input/state.rs#L226-L243

duanebester:
> image image
> The inline completion text has not render correct.

Yeah, this was on purpose as I didn't want to "shift" the existing text down, but it seems like early versions of vscode copilot suggestions end up extending the current line for multiple lines before the insert and thus didn't have to paint over existing text

duanebester:
> In Zed, there have a DiaplayMap for render, that is difference like the actual text. But it is a bit complex, I have not understood it all.
> 
> In this case, I think we must insert the inline completion text to the `text`. Or we have a holder (Maybe we can do like the password mask (I am not sure if it is possible).
> 
> ### For example:
> If we have added an empty (Only keep `\s`, and with same length of the inline completion text) placeholder into the source text before `shape_line`. After this, the primary text paint will leave a space for us to paint the inline completion.
> 
> But this you must handle more details, the `shape_line` will keep the `lines` in `LastLayout` for selection, position cursor on mouse click ..., we need to ensure if user accept or not of the inline completion, the `LastLayout` will have a correct info.
> 
> https://github.com/longbridge/gpui-component/blob/26ea1576bf744fbb0e5a6bdc234921121cd5722d/crates/ui/src/input/state.rs#L226-L243

I'll give this idea a shot! Overall, you think inline completions would be a nice feature to have?

huacnlee:
Yes, it's a very nice feature, if we can be done that will good.

I have read your code, most other part are all good, you can see I have formatted something.

So here we need to fix this last thing.

duanebester:
Recent changes:

1. **PrepaintState**: Add 2 new fields for inline completion ghost lines
2. **New helper method**: `layout_inline_completion()` shapes the completion text into first line + ghost lines
3. **`prepaint`**: Call the helper, adjust `scroll_size` to include ghost line height
4. **Text paint loop**: After painting cursor row, paint ghost lines and shift `offset_y`
5. **First line paint**: Paint completion's first line suffix after cursor position
6. **Line numbers loop**: Also shift `offset_y` after cursor row for proper alignment

duanebester:
https://github.com/user-attachments/assets/8e437f1e-9792-4417-b070-0a1dd8de6024

huacnlee:
Looks great, thank you.

huacnlee (approved):

PR #1767, opened 2025-12-08
text_view: Fix CodeBlock interactivity id

duanebester:
Use `ix` for TextView elements id to support CodeBlock actions to use short `id` name


huacnlee:
I will make change for this PR, to use `ix` for elements id in TextView.
GPUI has its own GloabalId design, we not need UUID.

huacnlee:
@duanebester Try my changed version

PR #1770, opened 2025-12-09
input: Add `context_menu` option to support disable ContextMenu.

pedrobartolini:
## Description

Popovers already use defer_draw to render above other content, thus right-clicking an input inside a popover (triggering a context menu which also uses defer_draw), you get nested deferred drawing → panic.

This small change allows users to disable the menu, making it safe to right-click inputs inside popovers.

## How to Test

```rust
InputState::new(window, cx)
   .default_value(format!("{}/{}", config.ipv4, config.ipv4_prefix))
   .placeholder("IPv4")
   .context_menu(false)  // disables ContextMenu input is now safe to use inside popovers
```

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -302,6 +302,8 @@ pub struct InputState {
     /// The mask pattern for formatting the input text
     pub(crate) mask_pattern: MaskPattern,
     pub(super) placeholder: SharedString,
+    /// Whether to show context menu. (default: true)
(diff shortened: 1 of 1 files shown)


huacnlee:
This not the best way to solve issue, let me think about.

Ref https://github.com/zed-industries/zed/issues/42765

huacnlee:
Close this by https://github.com/zed-industries/zed/pull/47770

PR #1791, opened 2025-12-11
pagination: Add Pagination component

18o:
## Screenshot

image

### DIFF
--- a/docs/docs/components/index.md
+++ b/docs/docs/components/index.md
@@ -20,6 +20,7 @@ collapsed: false
 - [Image](image) - Image display with fallbacks
 - [Kbd](kbd) - Keyboard shortcut display
 - [Label](label) - Text labels for form elements
+- [Pagination](pagination) - Page navigation controls
 - [Progress](progress) - Progress bars
 - [Radio](radio) - Single selection from multiple options
 - [Skeleton](skeleton) - Loading placeholders
--- a/docs/docs/components/pagination.md
+++ b/docs/docs/components/pagination.md
@@ -42,7 +42,9 @@ Pagination::new("my-pagination")
 
 ### Compact Style
 
-The compact style only shows the previous and next buttons with icons, without displaying page numbers:
+The compact style only shows the previous and next buttons with icons, without displaying page numbers.
+
+Use `compact` method to enable compact style:
 
 ```rust
 Pagination::new("my-pagination")


huacnlee:
I'd like this feature, but we should add full pagination first:

image

https://ui.shadcn.com/docs/components/pagination

18o:
> I'd like this feature, but we should add full pagination first:
> 
> image
> https://ui.shadcn.com/docs/components/pagination

ok i'll update it to match shadcn

huacnlee:
And I have reviewed your current API design. I think, we should need to consider more, and read the other exist components and reference the Shadcn's API.

crates/ui/src/pagination.rs
```diff
@@ -0,0 +1,198 @@
...
+    pub fn total_pages(mut self, pages: u32) -> Self {
+        self.total_pages = pages;
+        self
+    }
+
+    /// Set the total number of items across all pages.
+    pub fn total_items(mut self, items: i64) -> Self {
+        self.total_items = items;
+        self
+    }
```
huacnlee, line 68: This just used to display a total info. I think, it is unnecessary, unless you want use to calculate the pages.

crates/ui/src/pagination.rs
```diff
@@ -0,0 +1,198 @@
...
+    pub fn prev_icon(mut self, icon: IconName) -> Self {
+        self.prev_icon = icon;
+        self
+    }
+
+    /// Set a custom icon for the next button.
+    pub fn next_icon(mut self, icon: IconName) -> Self {
+        self.next_icon = icon;
+        self
+    }
```
huacnlee: The icon should be an internal, like other components. If users want to customize, the assets folder files can override.

crates/ui/src/pagination.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    /// Set custom information text instead of the default.
+    ///
+    /// If not set, displays "Page {current} of {total} • {items} items"
+    pub fn info_text(mut self, text: impl Into<SharedString>) -> Self {
+        self.info_text = Some(text.into());
+        self
+    }
```
huacnlee, line 87: This is not a good API, more like a hard code.

We can use `child` do something like this.

crates/ui/src/pagination.rs
```diff
@@ -0,0 +1,198 @@
...
+
+impl Styled for Pagination {
+    fn style(&mut self) -> &mut StyleRefinement {
+        &mut self.style
+    }
+}
+
+impl RenderOnce for Pagination {
+    fn render(mut self, _: &mut Window, cx: &mut App) -> impl IntoElement {
+        let muted_fg = cx.theme().muted_foreground;
```
huacnlee: Use `cx.theme().muted_foreground` in element tree directly. Avoid to make a var unless you have logic.

crates/ui/src/pagination.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        let info_text = self.info_text.unwrap_or_else(|| {
+            SharedString::from(format!(
+                "Page {} of {} • {} items",
+                self.current_page, self.total_pages, self.total_items
+            ))
+        });
+        let base_id = self.base.interactivity().element_id.clone();
+        self.base
+            .refine_style(&self.style)
```
huacnlee: Always placement the `refine_style` after the internal style, then users can override all the exist styles.

crates/ui/src/pagination.rs
```diff
@@ -0,0 +1,198 @@
...
+                        }
+                        btn
+                    })
+                    .child({
+                        let mut btn =
+                            Button::new(SharedString::from(format!("{:?}-next", base_id)))
+                                .icon(self.next_icon)
+                                .compact()
+                                .disabled(!can_next);
+                        if let Some(handler) = self.on_next {
```
huacnlee: Use `when`, to avoid have a `let mut btn =` here.

18o:
Thanks for pointing those out. I will update accordingly.

huacnlee:
Thank you.
I have made some changes, now it is ready to merge.

PR #1796, opened 2025-12-12
dialog: add `padding_top` attribute to avoid overlay mask cover the custom height title bar.

PleaseDont:


crates/ui/src/dialog.rs
```diff
@@ -231,6 +233,12 @@ impl Dialog {
         self
     }
 
+    /// Set the top offset of the overlay, defaults to None, will cover the window area.
+    pub fn padding_top(mut self, padding_top: Pixels) -> Self {
```
huacnlee, line 237: I don't like named of this.

crates/ui/src/dialog.rs
```diff
@@ -119,6 +120,7 @@ impl Dialog {
             footer: None,
             children: Vec::new(),
             margin_top: None,
+            padding_top: None,
```
huacnlee, line 123: This should default to TITLE_BAR_HEIGHT

crates/ui/src/dialog.rs
PleaseDont: > I don't like named of this.

How about overlay_margin_top?

crates/ui/src/dialog.rs
PleaseDont: > This should default to TITLE_BAR_HEIGHT

If using system title bar, this should be zero, and gpui-component's title bar it is TITLE_BAR_HEIGHT, or someone could have a custom height title bar.

PR #1797, opened 2025-12-12
dialog: add `overlay_top` for customing overlay margin top.

PleaseDont:


crates/ui/src/dialog.rs
```diff
@@ -231,6 +233,14 @@ impl Dialog {
         self
     }
 
+    /// Set the top offset of the overlay
+    ///
+    /// When not using [`TitleBar`] set your own value, for System title bar it should be `px(0.0)`.
+    pub fn overlay_top(mut self, overlay_top: Pixels) -> Self {
```
huacnlee: `impl Into<Pixels>`

PR #1816, opened 2025-12-16
theme: Impl parse tailwind style color expr.

PleaseDont:
## Description

- Implement parse tailwind style color expression for `theme::schema::try_parse_color`.
- Change impl `From<&str>` for `ColorName` to `TryFrom`.

One can use like `black`, `blue-300`, `green/80`, or `yellow-600/33` in the theme config file.


huacnlee:
Thank you, this is really useful.

PR #1834, opened 2025-12-19
notification: Add animation to Notification and support stack collapse.

huacnlee:
https://github.com/user-attachments/assets/43ec75e7-a396-43cc-a999-23d40e09aae3


huacnlee:
Close this incomplete implementation. 

There are still some animation details it unable to fix.

PR #1839, opened 2025-12-21
input: Update paint argument for GPUI API change.

18o:
## Description

Fix the input element break caused by PR https://github.com/zed-industries/zed/pull/45417


huacnlee:
Thanks

PR #1844, opened 2025-12-22
Added IDE like Autocomplete Suggestions for InputState

Nareshix:
3 new functions.
- **autocomplete_words()** - pass in a vector of strings, and it uses those strings as autocomplete suggestions.
- **autocompletion_menu_height()** - customise max height autocompletion menu.
- **autocompletion_menu_width()** - customise max width of autocompletion menu.

https://github.com/user-attachments/assets/826ead45-88ba-476c-99d3-38a132dd8d83

## Notes about performance
- Really fast. even at ~500k words, it takes at most 1 second to load the autocomplete suggestions 

## How to Test

copy paste this into main.rs. Added commments to show where the new functions are being used

```rust
use gpui::*;
use gpui_component::{
    ActiveTheme, Root, Theme, ThemeMode, TitleBar,
    input::{Input, InputState},
};

struct ShowcaseView {
    editor: Entity<InputState>,
}

impl ShowcaseView {

        // // if using linux u can use this instead of the 300 manual words below. this loads a large dictionary to test for performance
        // // it has close to about 500k words
        // let words: Vec<String> = std::fs::read_to_string("/usr/share/dict/words")
        //     .unwrap_or_else(|_| "hello world".to_string())
        //     .lines()
        //     .map(|s: &str| s.to_string())
        //     // .take(1000)  // Limit to first 1000 words. Remove this line to include all 500k words.
        //     .collect();

        let words = vec![
    "apple","river","silent","mountain","window","forest","cloud","amber","stone","early light",
    "paper","signal","orbit","memory","canvas","garden","theory","engine","bottle","clear sky",
    "planet","shadow","thread","mirror","signal flare","velocity","anchor","horizon","method","open field",
    "library","corner","pattern","texture","harbor","drift","volume","circuit","balance","cold rain",
    "system","object","motion","archive","border","syntax","packet","kernel","quiet room",
    "context","variant","buffer","network","signal loss","cursor","stream","vector","domain","wide angle",
    "history","future","present","random access","storage","compute","cluster","latency","throughput","soft limit",
    "window pane","process","thread pool","scheduler","runtime","compiler","linker","binary","source code",
    "integer","floating point","precision","overflow","underflow","error state","panic!","result type",
    "option value","borrow checker","ownership model","lifetime bound","unsafe block","mutable reference",
    "immutable data","stack frame","heap allocation","memory leak","race condition","deadlock",
    "semaphore","mutex lock","atomic value","cache line","branch predictor","instruction set",
    "pipeline","register file","clock cycle","power draw","thermal limit","cooling fan",
    "battery life","charging port","signal noise","frequency band","analog input","digital output",
    "sensor array","control loop","feedback signal","error margin","tolerance range","safety check",
    "failure mode","recovery path","fallback option","default value","user input","command line",
    "config file","environment variable","startup script","shutdown hook","log entry","debug trace",
    "error message","warning level","info channel","verbose mode","quiet flag","dry run",
    "test case","unit test","integration test","benchmark run","performance metric","latency spike",
    "throughput drop","resource usage","cpu load","memory pressure","disk io","network jitter",
    "packet loss","retry count","timeout window","backoff strategy","exponential delay","rate limit",
    "quota exceeded","access denied","permission error","authentication failed","authorization scope",
    "user profile","account settings","privacy mode","secure channel","encrypted payload","key exchange",
    "public key","private key","digital signature","certificate chain","trust anchor","root authority",
    "session token","refresh cycle","expiration time","clock skew","time drift","event timestamp",
    "ordered sequence","unordered set","hash map","tree structure","graph traversal","depth first",
    "breadth first","shortest path","weighted edge","node value","leaf node","root node",
    "balanced tree","rotation step","merge operation","split point","search index","query plan",
    "execution engine","cost estimate","optimizer hint","statistics table","cardinality guess","join strategy",
    "inner join","outer join","left side","right side","null value","constraint check",
    "foreign key","primary key","unique index","composite key","transaction log","commit record",
    "rollback segment","isolation level","serializable mode","snapshot read","write conflict","lock wait",
    "dead row","vacuum process","cleanup task","background worker","maintenance window","health check",
    "status page","uptime ratio","service level","error budget","incident report","postmortem note",
    "action item","follow up","design review","architecture diagram","data flow","control plane",
    "management layer","deployment target","release candidate","version bump","semantic version",
    "breaking change","migration step","compatibility layer","legacy mode","deprecation notice",
    "feature flag","rollout phase","canary release","full rollout","rollback plan","final approval", "hi", "there", "inline completion",
    "scrollable", "random word", "example text", "sample input", "test string", "demo content", "placeholder", "lorem ipsum",
];

        let editor = cx.new(|cx| {
            InputState::new(window, cx)
                .multi_line(true)
                .autocompletion_menu_width(1000.0, cx) // width
                .autocompletion_menu_height(350.0, cx)  //  height
                .autocomplete_words(words)  // suggestions
        });

        Self { editor }
    }
}

impl Render for ShowcaseView {
    fn render(&mut self, _: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
        div()
            .size_full()
            .bg(cx.theme().background)
            .child(TitleBar::new())
            .child(
                div().size_full().p_4().child(
                    div()
                        .size_full()
                        .border_1()
                        .border_color(cx.theme().border)
                        .bg(cx.theme().background)
                        .child(Input::new(&self.editor).h_full()),
                ),
            )
    }
}

fn main() {
    Application::new()
        .with_assets(gpui_component_assets::Assets)
        .run(|cx| {
            gpui_component::init(cx);
            cx.open_window(
                WindowOptions {
                    window_bounds: Some(WindowBounds::centered(size(px(1200.), px(800.)), cx)),
                    titlebar: Some(TitleBar::title_bar_options()),
                    ..Default::default()
                },
                |window, cx| {
                    Theme::change(ThemeMode::Dark, Some(window), cx);
                    let view = cx.new(|cx| ShowcaseView::new(window, cx));
                    cx.new(|cx| Root::new(view, window, cx))
                },
            )
            .unwrap();
        });
}

```

## Checklist

- [ x ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) document and followed the guidelines.
- [ x ] Reviewed the changes in this PR and confirmed AI generated code (If any) is accurate.
- [ x ] Passed `cargo run` for story tests related to the changes.
- [ x ] Tested macOS, Windows and Linux platforms performance (if the change is platform-specific)

### DIFF
--- a/crates/ui/src/input/lsp/completions.rs
+++ b/crates/ui/src/input/lsp/completions.rs
@@ -153,7 +153,13 @@ impl InputState {
         }
 
         if !matches!(self.context_menu, Some(ContextMenu::Completion(_))) {
-            let menu = CompletionMenu::new(cx.entity(), self.autocompletion_menu_width, window, cx);
+            let menu = CompletionMenu::new(
+                cx.entity(),
+                self.autocompletion_menu_width,
+                self.autocompletion_menu_height,
+                window,
+                cx,
+            );
             self.context_menu = Some(ContextMenu::Completion(menu));
         }
 
--- a/crates/ui/src/input/popovers/completion_menu.rs
+++ b/crates/ui/src/input/popovers/completion_menu.rs
@@ -229,6 +229,7 @@ pub struct CompletionMenu {
     query: SharedString,
     _subscriptions: Vec<Subscription>,
     max_width: Pixels,
+    max_height: Pixels,
 }
 
 impl CompletionMenu {
@@ -238,7 +239,7 @@ impl CompletionMenu {
     pub(crate) fn new(
         editor: Entity<InputState>,
         width: Pixels,
-
+        height: Pixels,
         window: &mut Window,
         cx: &mut App,
     ) -> Entity<Self> {
@@ -277,6 +278,7 @@ impl CompletionMenu {
                 query: SharedString::default(),
                 _subscriptions,
                 max_width: width,
+                max_height: height,
             }
         })
     }
@@ -358,6 +360,11 @@ impl CompletionMenu {
         self.list.update(cx, |list, _| {
             list.delegate_mut().max_width = width;
         });
+        cx.notify();
+    }
+
+    pub fn set_height(&mut self, height: Pixels, cx: &mut Context<Self>) {
+        self.max_height = height;
         cx.notify();
     }
 
@@ -470,7 +477,16 @@ impl Render for CompletionMenu {
         let window_size = window.bounds().size;
 
         let available_space_right = window_size.width - abs_pos.x - POPOVER_GAP;
-        let menu_width = self.max_width.min(available_space_right);
+        let menu_width = self
+            .max_width
+            .min(available_space_right)
+            .min(window_size.width * 0.8);
+
+        let available_space_bottom = window_size.height - abs_pos.y - POPOVER_GAP;
+        let menu_height = self
+            .max_height
+            .min(available_space_bottom)
+            .min(window_size.height * 0.8);
 
         let selected_documentation = self
             .list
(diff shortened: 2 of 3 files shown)


huacnlee:
I believe that the `autocomplete_words` function doesn't need to be implemented internally; the LSP's `auto_complete` function can achieve the same result. Everything you do can be accomplished by implementing `CompletionProvider`.

Here are already examples for how to write a `CompletionProvider` with multiple types of completion results.

```
cargo run --example editor
```

huacnlee:
I'm sorry, you have write so much code, but this PR cannot be accepted into the Editor.

PR #1846, opened 2025-12-23
notification: Add `NotificationSettings` for config Notification placement.

### LINKED ISSUE #1683: Add customizable position and offset options for Notification component
I’d like to propose two enhancements to the Notification component to improve its flexibility:

1、Customizable Position
Currently, notifications always appear in the top-right corner of the screen. It would be great if users could control the placement via a position prop with the following options:
`top-right`(default)
`top-left`
`bottom-right`
`bottom-left`

2、Offset Support
Allow specifying an `offset` (e.g., in pixels) to define the distance between the notification and the screen edge. This would help avoid overlapping with system UI elements (like macOS menu bar or Windows taskbar) or other app components.
These options are common in desktop notification systems and would significantly improve the component’s usability in different layout contexts.

Thanks for considering this suggestion!

PleaseDont:
Closes #1683 

## Screenshot

https://github.com/user-attachments/assets/68624373-4abe-4f99-ad94-4217c62005b0

https://github.com/user-attachments/assets/a2924fb1-d0cc-4a0f-b2af-7dfdd42a95dd

### DIFF
--- a/crates/ui/src/root.rs
+++ b/crates/ui/src/root.rs
@@ -117,7 +117,7 @@ impl Root {
             _ => (None, None, None, None),
         };
 
-        let placement = cx.theme().notification.placement.clone();
+        let placement = cx.theme().notification.placement;
 
         Some(
             div()
--- a/crates/story/src/stories/notification_story.rs
+++ b/crates/story/src/stories/notification_story.rs
@@ -82,7 +82,7 @@ impl Render for NotificationStory {
                         .outline()
                         .label(cx.theme().notification.placement.to_string())
                         .dropdown_menu(move |menu, window, cx| {
-                            let menu = ANCHORS.clone().into_iter().fold(menu, |menu, placement| {
+                            let menu = ANCHORS.into_iter().fold(menu, |menu, placement| {
                                 menu.item(
                                     PopupMenuItem::new(placement.to_string())
                                         .checked(cx.theme().notification.placement == placement)
--- a/crates/ui/src/notification.rs
+++ b/crates/ui/src/notification.rs
@@ -287,7 +287,7 @@ impl Render for Notification {
             Some(type_) => Some(type_.icon(cx)),
         };
         let has_icon = icon.is_some();
-        let placement = cx.theme().notification.placement.clone();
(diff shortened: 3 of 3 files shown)


crates/ui/src/root.rs
```diff
@@ -103,24 +103,71 @@ impl Root {
...
+
+    /// Render Notification layer with options.
+    ///
+    /// # Arguments
+    /// * `options` - `NotificationOptions` set the placement, paddings of the notification display, and layer_top avoid overlap with the custom title bar.
+    ///
+    pub fn render_notification_layer_options(
+        window: &mut Window,
+        cx: &mut App,
+        options: NotificationOptions,
```
huacnlee: 1. `render_notification_layer_options` -> `render_notification_layer_with_options`
2. Move `options` to as first argument. In GPUI style, we always keep `window`, `cx` at the end of the method (unless closure argument; keep the closure argument at the end).

huacnlee (changes requested):
I think we could refer to the `scrollbar_show` parameter and make `notification: NotificationOptions` a parameter of the Theme.

Then we do not need to always update the options here.

https://github.com/longbridge/gpui-component/blob/6cfdf7d683f83d229946e0419bae534627c6b15a/crates/ui/src/theme/mod.rs#L69

PleaseDont:
> Thanks, it's a good feature for the GPUI Component. But the API we need to change.
> 
> I think we could refer to the `scrollbar_show` (I will consider renaming it to `scrollbar: ScrollbarSetting` after this PR is done.) parameter and make `notification: NotificationSetting` a parameter of the Theme.
> 
> Then we do not need to always update the options here.
> 
> https://github.com/longbridge/gpui-component/blob/6cfdf7d683f83d229946e0419bae534627c6b15a/crates/ui/src/theme/mod.rs#L69

It's great, putting notification setting in the theme is indeed much more convenient, after all, an app won't always change the placement of notifications.
At the beginning, I was wondering if I wanted to draw a notification list for 6 placement simultaneously in root, but it was completely unnecessary.

huacnlee (dismissed):

huacnlee:
Thank you.

PR #1848, opened 2025-12-23
table: Fix `row_selectable` not going to limit row selection.

### LINKED ISSUE #1847: table:   Table row_selectable not use
TableState has row_selectable  but not use.

Image

I think it should be added here. I don't know if it's accurate.

Image

doubo6sir:
Closes #1847 

## Description

Add a row_selectable check in fn on_row_left_click.


huacnlee:
Thanks

PR #1861, opened 2025-12-27
notification: Change the close button layout from `absolute` to `v_flex`.

PleaseDont:
## Description

- Absolute layout may cover the content.
- Parent's padding top may changed.

## Screenshot

|  |  |


huacnlee:
Thanks for the PR, but I have make a better change in https://github.com/longbridge/gpui-component/pull/1872

PR #1864, opened 2025-12-28
Button: add support for custom text sizes in button component

18o:
## Description

Currently, the button component does not allow setting the label text size, even if specified within the button `.button_text_size(Size::XSmall)`, it does not take effect. This PR makes the font setting method work correctly.

## Screenshot

CleanShot20251228205330

CleanShot20251228205421

### DIFF
--- a/crates/story/src/stories/button_story.rs
+++ b/crates/story/src/stories/button_story.rs
@@ -913,7 +913,7 @@ impl Render for ButtonStory {
                         Button::new("button-custom-text-large")
                             .primary()
                             .label("Large")
-                            .button_text_size(Size::Large),
+                            .text_lg(),
                     )
                     .child(
                         Button::new("button-custom-text-medium")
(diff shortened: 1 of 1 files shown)


crates/ui/src/styled.rs
```diff
@@ -531,7 +531,9 @@ impl<T: Styled> StyleSized<T> for T {
         match size {
             Size::XSmall => self.text_xs(),
             Size::Small => self.text_sm(),
-            _ => self.text_base(),
+            Size::Medium => self.text_base(),
+            Size::Large => self.text_lg(),
```
huacnlee, line 535: Large and custom size button should keep the text base.

18o: What if I want to display small text inside a large button?

huacnlee:
Not the aim of the Button design. This customize case you can write a simple Button implement by use `div`.

PR #1871, opened 2025-12-30
tree: Add `set_selected_item` method.

### LINKED ISSUE #1869: Tree: add set_selected_index_by_id
we maintain entry `id` is easy, but `index` is very hard

yoogoc:
Closes #1869

### DIFF
--- a/crates/ui/src/tree.rs
+++ b/crates/ui/src/tree.rs
@@ -265,6 +265,12 @@ impl TreeState {
         cx.notify();
     }
 
+    /// Get the currently selected tree item, if any.
+    pub fn selected_item(&self) -> Option<&TreeItem> {
+        self.selected_ix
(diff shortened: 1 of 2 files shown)


crates/ui/src/tree.rs
```diff
@@ -228,6 +243,22 @@ impl TreeState {
         cx.notify();
     }
 
+    /// Set the selected index by entry id, or `None` to clear selection.
+    pub fn set_selected_by_id(&mut self, id: Option<SharedString>, cx: &mut Context<Self>) {
```
huacnlee: This API should be named `set_selected_item` and use `item: Option<&TreeItem>`,

huacnlee: And there also needs to be a `selected_item` to return a `Option<&TreeItem>`

huacnlee (approved):

PR #1896, opened 2026-01-05
sheet: Add support for `SheetSettings`

### LINKED ISSUE #1891: overlay_top removed means that sheets can't be used without titlebar
Sheet::overlay_top allowed for removing the extra space when a titlebar wasn't present.

However, https://github.com/longbridge/gpui-component/pull/1865 removed this method.

I don't understand the reason for the removal.

Did the author overlook the fact that not everyone uses a builtin titlebar?

madcodelife:
Close #1891.

| Left | Top | Right | Bottom |
| - | - | - | - |
| left | top | right | bottom |


huacnlee (approved):

huacnlee:
@Moulberry Checkout this new change is this solve your problem?

Moulberry:
@huacnlee I think so, I can update the theme inside the Root render() method. It's a little bit messy

PR #1916, opened 2026-01-09
table: Add more keybindings to Table.

### LINKED ISSUE #1547: Page Up & Page Down in table and other components
I opened the table example, and when I focus to the table, it is possible to navigate it with up and down arrow keys, but not with page up and page down.

Also if I hit TAB-key to change focus from the "Number of stocks" text field above the table, it won't change the focus to table below. 

It's only possible to focus the table rows by clicking with the mouse first and then navigating with arrows.

When keyboard focused to table, I would expect Page Up & Page Down to work, maybe Home and End keys as well?

Hizome:
Closes #1547

## Description
add key map to table component : 
- tab
- shift+tab
- PgUp
- PgDn
- Home
- End


Hizome:
It still need a mouse click to foucs first, I kind of dont know how to just use tab to foucs the table, give me some time for it

Hizome:
the tab focus already working

PR #1949, opened 2026-01-15
docs: Fix description list example

fhluo:
Fix `DescriptionList` example to use the correct `item` method.


madcodelife:
Thanks.

PR #1968, opened 2026-01-22
select: Fix `perform_search` to call `SelectItem::matches`

fhluo:
Previously, `SearchableVec::perform_search` used hardcoded title matching, which prevented users from implementing custom search logic.

### DIFF
--- a/crates/ui/src/select.rs
+++ b/crates/ui/src/select.rs
@@ -428,7 +428,7 @@ impl<I: SelectItem> SelectDelegate for SearchableVec<I> {
         self.matched_items = self
             .items
             .iter()
-            .filter(|item| item.title().to_lowercase().contains(&query.to_lowercase()))
+            .filter(|item| item.matches(query))
             .cloned()
             .collect();
 


huacnlee:
No, this is correct. 
All selection filter should be search by case insensitive.

And this also just for `SearchableVec` is a bulit-in implemention for Select. If there have some special reason not like this, users can implemention it own `SelectDelegate`.

fhluo:
Thanks for reviewing.

The default `SelectItem::matches` implementation already does case-insensitive matching. 
```rust
// crates/ui/src/select.rs:52-54
fn matches(&self, query: &str) -> bool {
    self.title().to_lowercase().contains(&query.to_lowercase())
}
```

This aligns with how `SearchableVec<SelectGroup<I>>` already works.
```rust
// crates/ui/src/select.rs:484-502
fn perform_search(
    &mut self,
    query: &str,
    _window: &mut Window,
    _: &mut Context<SelectState<Self>>,
) -> Task<()> {
    self.matched_items = self
        .items
        .iter()
        .filter(|item| item.matches(&query))
        .cloned()
        .map(|mut item| {
            item.items.retain(|item| item.matches(&query));
            item
        })
        .collect();
    Task::ready(())
}
```

huacnlee:
Thanks, now I see the point.

PR #1970, opened 2026-01-23
editor: Adding support of PHP and fix combined injection

PRRPCHT:
## Description

Adds PHP syntax highlighting to the code editor component using tree-sitter-php.

## Screenshot

Before:
before

After:
after

## How to Test

1. Run `cargo run -p gpui-component-story --example editor`
2. Open a PHP file
3. The syntax of the file is now highlighted

## Checklist

- [X ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) document and followed the guidelines.
- [ X] Reviewed the changes in this PR and confirmed AI generated code (If any) is accurate.
- [ X] Passed `cargo run` for story tests related to the changes.
- [ N/A] Tested macOS, Windows and Linux platforms performance (if the change is platform-specific)

### DIFF
--- a/crates/ui/src/highlighter/highlighter.rs
+++ b/crates/ui/src/highlighter/highlighter.rs
@@ -9,7 +9,7 @@ use std::{
     ops::Range,
     usize,
 };
-use tree_sitter::{InputEdit, Node, Parser, Point, Query, QueryCursor, StreamingIterator, Tree};
+use tree_sitter::{InputEdit, Parser, Point, Query, QueryCursor, StreamingIterator, Tree};
 
 /// A syntax highlighter that supports incremental parsing, multiline text,
 /// and caching of highlight results.


huacnlee:
image

Some tag highlight are not correct.

PRRPCHT:
Thanks for your feedback and the test file!

Indeed after investigations there's a case where the closing html tags are not highlighted. It looks like there's no trivial solution, I'll look more into this as the solution may be somewhere in tree-sitter-php.

A short term solution could be to remove the HTML support, as per default.

PRRPCHT:
It was a tricky one but we made it!

## Problem

Tree-sitter-php splits HTML content into separate `text` nodes around `<?php ?>` blocks. For example:

```
<title><?php echo $title; ?></title>
```

Produces three pieces: `text("<title>")` → PHP code → `text("</title>")`. Each `text` node was being parsed as an independent HTML fragment, so the HTML parser saw `</title>` in isolation — an orphan closing tag with no matching opener — and couldn't highlight it.

This is a well-known issue ([tree-sitter-php#123](https://github.com/tree-sitter/tree-sitter-php/issues/123), [nvim-treesitter#2565](https://github.com/nvim-treesitter/nvim-treesitter/issues/2565)).

## Fix

In the context of the PHP highlighter:

1. **`injections.scm`**: Added `(#set! injection.combined)` to the `text` node rule. This is the standard tree-sitter directive that tells the highlighter to merge all disjoint text node ranges before parsing them as HTML.
2. **`highlighter.rs`**: Implemented support for `injection.combined` in the highlighter (which was already started but unfinished):
   - **Before:** the highlighter would take each HTML fragment individually (e.g. just `</title>`). The HTML parser, seeing an orphan closing tag with no context, couldn't make sense of it.
   - **After:** at startup, injection patterns marked `combined` are split into a dedicated query. At highlight time, this query scans the entire document and collects all `text` node byte ranges. These ranges are passed to the HTML parser via `Parser::set_included_ranges`. The parser then reads the original source but *skips over* the PHP gaps, seeing one continuous HTML document (`<html>...<title>...[skip]...</title>...</html>`). Opening and closing tags match correctly, and the resulting byte offsets map straight back to the original file.

This solution was inspired by commented code already present in highlighter.rs (and partly reused) and the implementation of PHP highlighting in Zed. It is **not** limited to PHP & HTML.

## Screenshot

Screenshot 2026-02-01 at 23 41 57

huacnlee (dismissed):

huacnlee:
Thank you.

PR #1984, opened 2026-01-31
input: Allow insert on input state when disabled

### LINKED ISSUE #1983: Input(InputState): Should allow insert when disabled
## Description

When `Input` is set to disabled, it also sets the `InputState` to disabled which prevents `InputState::insert` from working. This was already mentioned before in #1350, but the PR that resolved it #1362  only fixed it for `InputState::set_value`. 

## Environment

- GPUI: [git: main]
- GPUI Component: [git: 35f55766f71ef30ef8dfbbcf4f647fef49a99a83]
- Platform: [macOS 26]

## Steps to Reproduce

1. Create input `InputState::new(&state).disabled(true)`
2. Call `state.insert(...)`
3. See the text doesn't update

## Expected

Text to be updated with the call to `InputState::insert` and for `Input::disabled` to allow affect the user's ability to modify the text. Essentially a readonly mode instead of completely disabled and immutable.

## Actual

Text did not update when calling `InputState::insert`

z-jxy:
Closes #1983

## Description

Set `InputState.disabled` to false during `InputState::insert` and restore before returning. 

## Break Changes

Users who relied on the text to not update when `InputState::insert` is called if the `Input` is disabled. Not sure why someone would be doing this though.

- Change 1

```diff
    pub fn insert(
        &mut self,
        text: impl Into<SharedString>,
        window: &mut Window,
        cx: &mut Context<Self>,
    ) {
+       let was_disabled = self.disabled;
+       self.disabled = false;
        let text: SharedString = text.into();
        let range_utf16 = self.range_to_utf16(&(self.cursor()..self.cursor()));
        self.replace_text_in_range_silent(Some(range_utf16), &text, window, cx);
        self.selected_range = (self.selected_range.end..self.selected_range.end).into();
+       self.disabled = was_disabled;
    }
```

## How to Test

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

## Checklist

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -619,10 +619,7 @@ impl InputState {
         cx: &mut Context<Self>,
     ) {
         self.history.ignore = true;
-        let was_disabled = self.disabled;
-        self.disabled = false;
         self.replace_text(value, window, cx);
-        self.disabled = was_disabled;
         self.history.ignore = false;
 
         // Ensure cursor to start when set text


crates/ui/src/input/state.rs
huacnlee: Here need remove now.

z-jxy: Missed this, its removed now

PR #2002, opened 2026-02-06
Add `Iconify` component.

PleaseDont:
Manually download and collect icon resources is a little bit boring.

Add a new component `Iconify` that download svg from [Iconify](https://iconify.design) through Iconify API and cache them locally to embed in binary.


huacnlee:
This should be an external project to do.

PR #2011, opened 2026-02-10
editor: Add Kotlin syntax highlighting

PRRPCHT:
## Description

Add support for Kotlin in the code editor. This led to add 2 improvements in the tree-sitter implementation:

1. The highlighter merged distant captures with the same name into giant ranges which made in the code below the syntax highlighting in the coroutine disappear:

```kotlin
fun main() {
    println("this one is highlighted")        // highlighted

    coroutineScope {
        launch {
            println("this one is NOT")        // not highlighted
        }
    }
}
```
 Problem:
  - The highlighter merged distant captures with the same name into giant ranges
  - Example from above: coroutineScope and println both captured as @function got merged into a single giant range
  - This giant range was then overwritten by a generic @variable capture, causing println to lose its function highlighting

Solved by allowing only exact-overlap deduplication in highlighter.rs, line 474.

2. Predicate-based queries like #eq? @keyword "where" silently failed, even for valid matches

Problem:
  - The iterator skipped the first chunk when reading node text for predicate evaluation (called cursor.next() immediately).  If the node fits entirely within one chunk the the iterator returns nothing.
  - It returned entire chunks (potentially kilobytes) instead of slicing to just the node's bytes

Solved by:
- changes in the ByteChunks struct to remember where the node starts and ends, as well as if it's the first chunk
- change the next() function in highlighter.rs so that we don't ignore the first chunk and return only smaller slices of chunk

## Screenshot

kotlin

## How to Test

1. cargo run -p gpui-component-story --example editor
2. open test.kt

## Checklist

- [N/A] Tested macOS, Windows and Linux platforms performance (if the change is platform-specific)


huacnlee:
Thanks

PR #2059, opened 2026-02-22
theme: Add Oklch color type and default-colors use `oklch`.

huacnlee:


huacnlee:
Close this, a bit complex changes. 

I don't want introduce this for now.

PR #2096, opened 2026-02-26
sheet: Fix window drag not working when sheet is opened

### LINKED ISSUE #2046: Can not drag window when sheet is opened
## Description

Seems like dragging and interaction with TitleBar is not working when sheet is opened. When I click in titlebar area sheet is not closed, but TitleBar is not responsive.

## Environment

- GPUI: latest main branch
- GPUI Component: latest main branch
- Platform: Windows 11

mengh04:
The window drag functionality was not working when sheet was opened because window_control_area was only set for Placement::Bottom.

This change removes the placement check and always sets window_control_area
when overlay is enabled, making the window draggable in all sheet positions.

Fixes #2046

## Screenshot

| before | after |


madcodelife:
Thank you.

PR #2112, opened 2026-03-03
text: Fix correct TextView selection when dragging back to the anchor line

### LINKED ISSUE #2111: TextView selection can expand to line end after dragging back to the anchor line
## Description

When selecting text in `TextView`, dragging from one line to another line and then moving back to the original line can produce an incorrect selection range on the original line.

If the selection starts at around character 10 on a line, then the cursor moves to the next line, and finally returns to the original line:
- dragging back to around character 6 may select from character 6 to the end of the line instead of character 6 to character 10;
- dragging back to around character 15 may sometimes select from character 10 to the end of the line, and sometimes from character 15 to the end of the line.

This makes line selection unstable when the drag path temporarily crosses another line.

## Environment

- GPUI: `0.2.2` from `zed` git `42ba961075b16aaf35d48631f0ce4e1a4196d983`
- GPUI Component: `0.5.1`
- Platform: `macOS 26 / Darwin 25.3.0 / arm64`

## Steps to Reproduce

1. Render a selectable `TextView` with at least two lines of plain text.
2. Mouse down on the first line around character 10.
3. Drag to the next line.
4. Move the cursor back to the first line around character 6 or character 15.
5. Observe the highlighted range on the first line.

## Expected

When the cursor returns to the original line, the highlighted range on that line should be bounded by the original anchor position and the current cursor position.

https://github.com/user-attachments/assets/338182c7-c755-41ae-a48a-62d7f8ff451b

## Actual

When the cursor returns to the original line, `TextView` can treat the line as if it is still part of a cross-line selection and extend the selection to the line end instead of clamping it to the anchor/current positions.

suxiaoshao:
Closes #2111

## Description

Fix `TextView` selection when dragging back to the original line after crossing another line. The selection helper now treats points within the same rendered line as a single-line selection, so the highlight stays bounded by the anchor and current cursor positions. This also adds regression tests for the same visual line with different `y` positions.

AI-generated portions reviewed for accuracy: the selection helper adjustment and the regression tests in `crates/ui/src/text/inline.rs`.

### Before

https://github.com/user-attachments/assets/338182c7-c755-41ae-a48a-62d7f8ff451b

### After

https://github.com/user-attachments/assets/63c0f53c-4f56-4b0d-b3e5-da8bbfa4be58

## How to Test

1. Run `cargo test -p gpui-component test_point_in_text_selection --lib`.
2. In a selectable `TextView`, start a selection on one line, drag to the next line, then move back to the original line.
3. Verify the selection on the original line is bounded by the anchor position and the current cursor position instead of extending to the line end.

## Checklist

### DIFF
--- a/crates/ui/src/text/inline.rs
+++ b/crates/ui/src/text/inline.rs
@@ -380,6 +380,7 @@ fn point_in_text_selection(
     selection_end: Point<Pixels>,
     line_height: Pixels,
 ) -> bool {
+    let point_in_line = |point: Point<Pixels>| point.y >= pos.y && point.y < pos.y + line_height;
     let top = selection_start.y.min(selection_end.y);
     let bottom = selection_start.y.max(selection_end.y);
     let x = pos.x + char_width.half();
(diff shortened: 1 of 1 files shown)


huacnlee:
Thank you.

PR #2114, opened 2026-03-03
sheet: Fix losing focus when sheet is opened twice

Moulberry:
## Description

Fixes a bug where if open_sheet_at is called twice, then when the sheet is closed, the previous_focused_handle will be invalid. This is because the 2nd sheet uses the focus_handle of the previous sheet, which is no longer valid.
This PR fixes the issue by making the 2nd sheet inherit the previous_focused_handle from the currently active sheet.

## How to Test

I tested by adding a keybind for opening a sheet twice. Without this PR, the focus broke and required calling window.close_sheet before window.open_sheet_at.

## Checklist

### DIFF
--- a/crates/ui/src/root.rs
+++ b/crates/ui/src/root.rs
@@ -302,7 +302,11 @@ impl Root {
     ) where
         F: Fn(Sheet, &mut Window, &mut App) -> Sheet + 'static,
     {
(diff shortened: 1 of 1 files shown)


madcodelife:
Thank you.

PR #2128, opened 2026-03-09
editor: Improve highlighting performance

zanmato:
## Description

The editor generally works great when the code is valid. But when editing, and the code becomes invalid, the editor can become really slow because of how tree-sitter is trying to traverse way too many nodes. I notice this when I'm copying thousands of rows in my SQL-editor and attempting to edit the query.

This PR introduces a few improvements to highlighting,

1. Update the tree in a background task if it takes too long to do it synchronously (usually broken code). Perhaps we can put the parser in an `Arc` but I don't think it's necessary
2. Group contiguous ranges instead of calling `styles` for each line
3. Improve tree-sitter querying similar to what you've already done for folding (collect nodes), where we attempt to skip nodes that are outside of the visible range

tree-sitter parses my broken SQL to this program:
```
[program][1, 0] - [10003, 0]
  [ERROR] [1, 0] - [10001, 11]
    [ERROR] [1, 1] - [1, 6]
    [ERROR] [2, 1] - [10001, 6]
      [keyword_as] [10001, 2] - [10001, 4]
 ```
Which is quite problematic unless we try to bail parsing early.

Unfortunately I had to include all three improvements to get my broken SQL working :smile: 
There are more improvements to be made for editing large files, but we can leave that for another PR.

With these improvements we can even open 1 million line files (*but not really edit, that requires more improvements, `samply` says the time spent is in calculate_visible_range and layout line numbers*).

## Screenshot

**NOTE: I disabled the color lsp in these videos, because it is too slow for the large files**

### Before

https://github.com/user-attachments/assets/4620efdd-9bcf-4d51-8f43-6f20cea9ab1f

### After

https://github.com/user-attachments/assets/f41601b7-6ebb-47a2-bce0-8e3772a43ced

## How to Test

Generate broken sql, and big json files:
```sh
$ (echo ""; for i in $(seq 1 9999); do echo "('$i'),"; done; echo "('10000')"; echo ') as x(ix);') > test_broken.sql
$ (echo '['; for i in $(seq 1 999999); do echo "{\"id\": $i, \"name\": \"user_$i\"},"; done; echo '{"id": 1000000, "name": "user_1000000"}'; echo ']') > big_test.json
```

Try to open them in the editor

## Checklist

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -1062,7 +1062,7 @@ impl TextElement {
                     flush_range(start, line - 1, false, &mut styles);
                 }
 
-                flush_range(line, line + 1, true, &mut styles);
+                flush_range(line, line, true, &mut styles);
                 continue;
             }
 


zanmato:
Also, I think if we update tree-sitter to the latest version, then it will compile for WASM so we don't need to stub it, not sure about languages though

zanmato:
The remaining issue for editing large-ish files (but small files will also benefit) is that in `layout_cursor` and `calculate_visible_range` we are always traversing from 0 ix which yields a lot of calls to `visible_wrap_row_count_for_buffer_line`. We need some clever way to estimate the starting index, based on scroll top maybe?

huacnlee:
> Also, I think if we update tree-sitter to the latest version, then it will compile for WASM so we don't need to stub it, not sure about languages though

This we can do later.

Your change too mush, I need time to test.

zanmato:
> > Also, I think if we update tree-sitter to the latest version, then it will compile for WASM so we don't need to stub it, not sure about languages though
> 
> This we can do later.
> 
> Your change too mush, I need time to test.

Yeah, this should be a separate PR

zanmato:
I just noticed it's allocating way too much memory for every edit, probably because of the text/tree being duplicated. **EDIT** Seems to be `lint_document` in the editor that is leaking quite a lot of memory when I traced it with heaptrack. This can be left for another PR

huacnlee:
### Code review

Found 2 issues:

1. `set_highlighter()` and `reset_highlighter()` do not cancel the background `parse_task` introduced by this PR. When either method is called (e.g., on language switch), only `highlighter` is cleared, but the in-flight background parse keeps running. On completion it calls `cx.notify()`, causing a spurious re-render. If `set_highlighter()` is called twice in quick succession, both background parses run concurrently with the old language. The sync-success path in `update_highlighter()` already does `parse_task.borrow_mut().take()` to cancel pending parses — the same pattern should be applied in these two methods.

https://github.com/longbridge/gpui-component/blob/6e71f22046eef0e15f64cff330382b44a049321e/crates/ui/src/input/state.rs#L575-L602

2. `apply_background_tree()` calls `parse_combined_injections()` on the main thread. The background task in `dispatch_background_parse` correctly offloads tree-sitter parsing to `background_executor`, but then `apply_background_tree` (called in the `spawn_in` window context) synchronously re-parses all injection layers on the main thread. For languages with heavy injections (PHP, HTML with embedded JS/CSS), this blocks the UI thread — the same problem the background-parse design was trying to avoid.

https://github.com/longbridge/gpui-component/blob/6e71f22046eef0e15f64cff330382b44a049321e/crates/ui/src/highlighter/highlighter.rs#L452-L460

https://github.com/longbridge/gpui-component/blob/6e71f22046eef0e15f64cff330382b44a049321e/crates/ui/src/input/state.rs#L2078-L2114

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

zanmato:
I tried with a big html+php file after the latest changes, and I found another issue in `unique_styles`, it really struggles with long lines, especially with soft wrap enabled. Try this file and scroll down a bit, around line 1000,
`wget -qO- https://en.wikipedia.org/wiki/Sweden | sed 's/title="/data-php="<?php echo '"'"'hello world'"'"'; ?>" title="/g' > wiki_test.php`

Just `wget -qO longbridge.html https://longbridge.com/sg` is even worse, because it's really long lines

zanmato:
I added skipping too long lines. This is the same behavior as vscode. Zed seems slightly more sophisticated and forces wrapping of long lines.

**Editor with longbridge.html**
image

**vscode with longbridge.html**
image

The php file is still a bit slow, but I have a very improved `unique_styles` that can be added later

zanmato:
Thanks for the help with the additional changes!

PR #2130, opened 2026-03-10
tab: Don't wrap tab text

Moulberry:
## Description

Prevent text wrapping in tab with whitespace_nowrap().
A user of my application showed the following screenshot, where the tab text is wrapping improperly:
image

While I wasn't able to replicate the issue (it seems quite rare), it should be fixable by setting whitespace_nowrap on the tab.

I also removed one overflow_hidden() because there was a 2nd one below it.

## How to Test

Unfortunately I could not replicate, but the change is very simple so it should be fine.


crates/ui/src/tab/tab.rs
```diff
@@ -615,7 +615,6 @@ impl RenderOnce for Tab {
             .gap_1()
             .items_center()
             .flex_shrink_0()
-            .overflow_hidden()
```
huacnlee, line 618: Why remove this?

Moulberry: > I also removed one overflow_hidden() because there was a 2nd one below it.

Image

huacnlee:
Thanks

PR #2135, opened 2026-03-11
theme: Update input background to match Shadcn style.

Moulberry:
## Description

Update the visuals of input/checkbox/select/etc. to match shadcn ui (https://ui.shadcn.com/docs/components/radix/input)

I believe this change is important for accessibility and usability because it makes input elements easier to see because they now have a different color to the background. It also makes it clear to the user that the inputs can be clicked/typed in.

## Screenshot

| image | image |

## How to Test

View the following stories:
Checkbox
Form
Input
OtpInput
List
Radio
Select
DatePicker

## Checklist

### DIFF
--- a/crates/ui/src/list/list.rs
+++ b/crates/ui/src/list/list.rs
@@ -1,5 +1,5 @@
-use instant::Duration;
 use std::ops::Range;
+use instant::Duration;
 
 use crate::actions::{Cancel, Confirm, SelectDown, SelectUp};
 use crate::input::InputState;
@@ -287,10 +287,7 @@ where
                     });
 
                     // Always wait 100ms to avoid flicker
-                    window
-                        .background_executor()
-                        .timer(Duration::from_millis(100))
-                        .await;
+                    window.background_executor().timer(Duration::from_millis(100)).await;
                     _ = this.update_in(window, |this, window, cx| {
                         this.set_searching(false, window, cx);
                     });
@@ -643,7 +640,6 @@ where
                             Size::Small => this.px_1p5(),
                             _ => this.px_2(),
                         })
-                        .bg(cx.theme().background)
                         .border_b_1()
                         .border_color(cx.theme().border)
                         .child(


crates/ui/src/theme/theme_color.rs
```diff
@@ -63,6 +63,8 @@ pub struct ThemeColor {
     pub info_foreground: Hsla,
     /// Info hover background color.
     pub info_hover: Hsla,
+    /// Background color for inputs such as Input, Select, etc.
+    pub input_background: Hsla,
```
huacnlee: We should not to add this.

There have `mix_oklab` method can use like the Shadcn CSS:

```
@supports (color:color-mix(in lab, red, red)) {
    .dark\:bg-input\/30:is(.dark *) {
        background-color: 
color-mix(in oklab, var(--input) 30%, transparent);
    }
}
```

In this PR https://github.com/longbridge/gpui-component/pull/2099 have example for use `mix_oklab` method. We should not add a new variable into theme. Just use `mix_oklab` in Input to get a special color for `background`.

But even that, we need avoid this effect other themes.

Moulberry: I now calculate `cx.theme().input_background()` as:

Dark:
`self.primary.mix_oklab(self.transparent, 0.045)`

Light:
`self.primary.mix_oklab(self.background, 0.045).opacity(0.3)`

0.045 is from shadcn (30% * 15%)

I use separate calculations for dark and light theme, because otherwise the light theme inputs become too dark. Shadcn also has different logic for dark/light, so I think this is ok

I tested with all the themes in the story, I didn't see any major problems.

Moulberry:
Here is a video showing the new calculation with all the themes

https://github.com/user-attachments/assets/6bc7a427-2376-426e-b95d-93b7ced55bd7

crates/ui/src/theme/theme_color.rs
huacnlee: No, you need remove that method. I don't want have `theme().input_background` API.

crates/ui/src/theme/mod.rs
```diff
@@ -179,13 +179,22 @@ impl Theme {
...
     pub(crate) fn editor_background(&self) -> Hsla {
         self.highlight_theme
             .style
             .editor_background
-            .unwrap_or(self.background)
+            .unwrap_or(self.input_background())
+    }
+
+    #[inline]
+    pub fn input_background(&self) -> Hsla {
```
huacnlee: Or just let this is `pub(crate)`

huacnlee: Or better move this to `input`, this is a `input` special.

crates/ui/src/theme/mod.rs
```diff
@@ -179,13 +179,22 @@ impl Theme {
...
 
-    /// Get the editor background color, if not set, use the theme background color.
+    /// Get the editor background color, if not set, use the input background color.
     #[inline]
     pub(crate) fn editor_background(&self) -> Hsla {
         self.highlight_theme
             .style
             .editor_background
-            .unwrap_or(self.background)
+            .unwrap_or(self.input_background())
```
huacnlee: I think we not need change this part.

huacnlee:
Let me take to change those issues.

PR #2136, opened 2026-03-11
editor: Faster `unique_styles`

zanmato:
## Description

The old method was O(n^2) with the styles loop inside of the intervals loop. This changes it to O(n log n) + an additional loop over the styles. ~60x improvement on really long lines which is where it had trouble.

```
OLD: styles on 63000 byte nested HTML took 2.948660759s, highlights 32286
NEW: styles on 63000 byte nested HTML took 49.565413ms, highlights 32286
```

## How to Test

I tested it with this in `highlighter.rs`. But I didn't include it in the PR because it's weird to assert on benchmarks.

```rust
#[test]
#[cfg(feature = "tree-sitter-languages")]
fn test_long_lines() {
    // Generate alot of nested HTML tags
    let tags = ["div", "p", "span", "strong", "i", "u"];

    let mut html = String::new();
    for _ in 0..1000 {
        for tag in &tags {
            html.push_str(&format!("<{tag}>"));
        }
        html.push_str("x");
        for tag in tags.iter().rev() {
            html.push_str(&format!("</{tag}>"));
        }
    }

    let rope = Rope::from_str(&html);
    let mut highlighter = SyntaxHighlighter::new("html");
    highlighter.update(None, &rope);

    let highlight_theme = HighlightTheme::default_light();
    let start = std::time::Instant::now();
    let range = 6500..63000;
    let highlights = highlighter.styles(&range, &highlight_theme);
    let elapsed = start.elapsed();

    eprintln!(
        "styles on {} byte nested HTML took {:?}, highlights {}",
        html.len(),
        elapsed,
        highlights.len(),
    );

    assert!(!highlights.is_empty(), "Should produce highlights");
}
```

## Checklist

### DIFF
--- a/crates/ui/src/highlighter/highlighter.rs
+++ b/crates/ui/src/highlighter/highlighter.rs
@@ -665,52 +665,57 @@ pub(crate) fn unique_styles(
         return styles;
     }
 
-    let mut intervals = BTreeSet::new();
-    let mut significant_intervals = BTreeSet::new();
-
-    // For example
-    //
-    // from: [(6..11), (6..11), (11..17), (17..25), (16..19), (25..59))]
-    // to:   [6, 11, 16, 17, 19, 25, 59]
-    intervals.insert(total_range.start);
-    intervals.insert(total_range.end);
-    for (range, _) in &styles {
-        intervals.insert(range.start);
-        intervals.insert(range.end);
-        significant_intervals.insert(range.end); // End points are significant for merging decisions
+    // Create intervals: (position, is_start, style_index)
+    let mut intervals: Vec<(usize, bool, usize)> = Vec::with_capacity(styles.len() * 2 + 2);
+    for (i, (range, _)) in styles.iter().enumerate() {
+        intervals.push((range.start, true, i));
+        intervals.push((range.end, false, i));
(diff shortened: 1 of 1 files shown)


crates/ui/src/highlighter/highlighter.rs
```diff
@@ -665,52 +665,57 @@ pub(crate) fn unique_styles(
...
-    let mut significant_intervals = BTreeSet::new();
-
-    // For example
-    //
-    // from: [(6..11), (6..11), (11..17), (17..25), (16..19), (25..59))]
-    // to:   [6, 11, 16, 17, 19, 25, 59]
-    intervals.insert(total_range.start);
-    intervals.insert(total_range.end);
+    // Create intervals: (position, is_start, style_index)
+    let mut intervals: Vec<(usize, bool, usize)> = Vec::with_capacity(styles.len() * 2 + 2);
```
huacnlee, line 669: I have test the new code by start editor example, the highlight is looks correct.

But, why here is `* 2 + 2`?

huacnlee: Oh, now I get it.

PR #2142, opened 2026-03-12
chore: Update GPUI and fix type mismatch

notruri:
## Description

In [scrollbar.rs](crates/ui/src/scroll/scrollbar.rs), it was adding `Point<Pixels>` offset to a `Size<Pixels>` viewport/content size directly. Newer gpui no longer accepts that implicit mixing.

- convert `Size<Pixels>` to `Point<Pixels>` before adding to `max_offset()`. https://github.com/longbridge/gpui-component/blob/9d5adc20cb739d810a76b1efc98f63ac2e07b23c/crates/ui/src/scroll/scrollbar.rs#L82
- convert the resulting `Point<Pixels>` back to `Size<Pixels>` for the `content_size()` return value. https://github.com/longbridge/gpui-component/blob/9d5adc20cb739d810a76b1efc98f63ac2e07b23c/crates/ui/src/scroll/scrollbar.rs#L97
- convert `max_offset_for_scrollbar()` from `Point<Pixels>` to `Size<Pixels>` before adding it to `viewport_bounds().size`. https://github.com/longbridge/gpui-component/blob/9d5adc20cb739d810a76b1efc98f63ac2e07b23c/crates/ui/src/scroll/scrollbar.rs#L111

Also updated `gpui` and its components to pass the build, but I verified that nothing broke so far (unless I missed something), current gpui revision is at [f627c43](https://github.com/zed-industries/zed/commit/f627c43ea1e4a8dc5788b2136b7c78aedb6b87d3)

## How to Test

- [x] `cargo check`
- [x] `cargo clippy`
- [x] `cargo test`
- [x] `cargo run`

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -646,7 +646,7 @@ dependencies = [
  "bitflags 2.11.0",
  "cexpr",
  "clang-sys",
- "itertools 0.11.0",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -1122,7 +1122,7 @@ dependencies = [
 [[package]]
 name = "collections"
 version = "0.1.0"
-source = "git+https://github.com/zed-industries/zed#42ba961075b16aaf35d48631f0ce4e1a4196d983"
+source = "git+https://github.com/zed-industries/zed#e0881e38f91b87623795208615ca466415d1970e"
 dependencies = [
  "indexmap",
  "rustc-hash 2.1.1",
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -28,11 +28,28 @@ gpui-component-macros = { path = "crates/macros", version = "0.5.1" }
...
 story = { path = "crates/story" }
 
-gpui = { git = "https://github.com/zed-industries/zed" }
-gpui_platform = { git = "https://github.com/zed-industries/zed", features = ["font-kit", "x11", "wayland", "runtime_shaders"] }
-gpui_web = { git = "https://github.com/zed-industries/zed" }
-gpui_macros = { git = "https://github.com/zed-industries/zed" }
-reqwest_client = { git = "https://github.com/zed-industries/zed" }
+gpui = {
+    git = "https://github.com/zed-industries/zed",
+    rev = "f627c43ea1e4a8dc5788b2136b7c78aedb6b87d3"
```
huacnlee: Do not change this file, just cargo update

PR #2145, opened 2026-03-12
Use subtle background for inline code instead of theme accent

KlausUllrich:
## Summary

Replaces the inline code background color from `cx.theme().accent` to a subtle semi-transparent white overlay, reducing visual noise in markdown content.

## The Problem

Inline code spans (e.g., `some_function()` in a paragraph) currently use `cx.theme().accent` as their background highlight. The accent color is typically saturated and designed to draw attention — great for buttons and interactive elements, but too visually heavy for inline code mixed with regular text. The result is that inline code highlights dominate the reading flow instead of blending in.

## The Fix

Replace the accent background with a neutral semi-transparent overlay:

```rust
// Before
highlight.background_color = Some(cx.theme().accent);

// After
highlight.background_color = Some(gpui::Hsla { h: 0.0, s: 0.0, l: 1.0, a: 0.08 });
```

The `Hsla { h: 0.0, s: 0.0, l: 1.0, a: 0.08 }` value creates a very light white overlay (8% opacity) that:
- Provides enough contrast to distinguish inline code from surrounding text
- Works across light and dark themes (low alpha adapts to the underlying background)
- Doesn't compete with syntax highlighting in code blocks or other colored elements

## Context

Discovered while using gpui-component for markdown rendering in a desktop app. The change is minimal — one line in `crates/ui/src/text/node.rs`, in the `Paragraph::render` method where inline styles are applied. `cargo check` passes cleanly.

## Note

This is an opinionated visual change. An alternative approach would be to add a dedicated `code_inline_background` field to the theme/style system, giving users control. Happy to discuss if you'd prefer that direction.

### DIFF
--- a/crates/ui/src/text/node.rs
+++ b/crates/ui/src/text/node.rs
@@ -698,7 +698,11 @@ impl Paragraph {
                         });
                     }
                     if style.code {
-                        highlight.background_color = Some(cx.theme().accent);
+                        // Use a subtle semi-transparent overlay instead of the
+                        // bright accent color — inline code should blend with
+                        // surrounding text, not compete for attention.
+                        highlight.background_color =
+                            Some(gpui::Hsla { h: 0.0, s: 0.0, l: 1.0, a: 0.08 });
                     }
 
                     if let Some(mut link_mark) = style.link.clone() {


KlausUllrich:
Closing this — on reflection, replacing a theme-aware value with a hardcoded color isn't the right approach for upstream. The proper fix would be adding a dedicated style field for inline code backgrounds. Apologies for the noise.

PR #2202, opened 2026-03-30
[codex] Fix markdown table column layout

### LINKED ISSUE #2197: markdown table cells do not wrap and can overflow horizontally
## Description

Markdown table cells can overflow horizontally instead of wrapping inside the cell.

I hit this in an app that renders markdown via `TextView::markdown(...)`. Normal paragraph text wraps correctly, but once the content is inside a markdown table cell, long Chinese/English mixed text stays on a single line and pushes the row beyond the container width.

This looks like a bug in the markdown table renderer in `crates/ui/src/text/node.rs`, not in app-level usage.

## Environment

- GPUI: v0.2.2
- GPUI Component: v0.5.1
- Platform: macOS

## Steps to Reproduce

1. Render markdown with `TextView::markdown(...)`.
2. Include a table with one short first column and one long second column.
3. Put long natural-language text in the second column.
4. View the rendered table in a constrained-width container.

Example markdown:

```md
| Framework | Detection Pattern |
| --- | --- |
| React | useEffect without cleanup return, addEventListener without removeEventListener, setInterval/setTimeout without cleanup, subscribe() without unsubscribe() |
| Vue | onMounted without matching onUnmounted cleanup, watch/watchEffect without stop handle, manual addEventListener inside setup without removal |
```

## Screenshots

Image

## Expected

Table cells should wrap text within the available cell width, similar to normal paragraph rendering.

## Actual

The table renderer sets a width using `relative(len as f32)` where `len` is derived from text length, and also applies `.truncate()` on each cell. This makes cells behave like fixed/non-wrapping content and causes horizontal overflow for long cell text.

Relevant code:

- `crates/ui/src/text/node.rs`
- `div().w(Length::Definite(relative(len as f32)))`
- `.truncate()`

From inspection:

- `.truncate()` disables wrapping by applying `whitespace_nowrap`
- `relative(...)` expects a fraction of parent width, but here it receives character-count-like values (`len as f32`), not a `0.0..=1.0` ratio

So the issue appears to be in `gpui-component`'s markdown table layout logic.

suxiaoshao:
Closes #2197

## Description

This PR fixes markdown table cells overflowing or wrapping incorrectly in `TextView::markdown(...)`.

The previous table renderer approximated column widths with content length and then relied on flex layout. That diverged from browser table layout behavior in two important ways:

1. text length is not a reliable proxy for rendered width, especially for mixed CJK / ASCII / emoji content
2. letting flex shrink columns independently does not match browser-style table column allocation, so short columns could wrap unexpectedly while long columns failed to absorb the available width correctly

The fix follows the browser `table-layout: auto` model more closely:

- measure each column's rendered `preferred` width from actual shaped text
- measure each column's `min-content` width from the longest unbreakable segment in its cells
- allocate final column widths against the table's actual available width
- when space is limited, shrink only the columns that can shrink, down to their measured minimum width
- when extra space is available, distribute it across columns proportionally from their preferred widths
- render cells with explicit final widths so the text layout receives stable width constraints

This keeps short columns stable, allows long columns to wrap first, and matches browser table behavior much more closely for both two-column and multi-column markdown tables.

This work was investigated and implemented with assistance from Codex GPT-5.4.

## Screenshot

| Before | After |
| --- | --- |
| Image | image |

## How to Test

1. Render markdown with a table that has a short first column and a long second column in a constrained-width container.
2. Confirm that the short header or short body cells do not wrap prematurely.
3. Confirm that the long column wraps within the table width instead of overflowing horizontally.
4. Render a three-column table where the middle column is long and the last column is short.
5. Confirm that the middle column shrinks and wraps before the short last column is compressed.

Validated locally with:

- `cargo test -p gpui-component table_cell_segments --lib`
- `cargo test -p gpui-component allocate_widths --lib`

## Checklist

### DIFF
--- a/crates/ui/src/text/node.rs
+++ b/crates/ui/src/text/node.rs
@@ -6,15 +6,16 @@ use std::{
 
 use gpui::{
     AnyElement, App, DefiniteLength, Div, ElementId, FontStyle, FontWeight, Half, HighlightStyle,
-    InteractiveElement as _, IntoElement, Length, ObjectFit, ParentElement, SharedString,
-    SharedUri, StatefulInteractiveElement, Styled, StyledImage as _, Window, div, img,
+    InteractiveElement as _, IntoElement, ObjectFit, ParentElement, SharedString, SharedUri,
+    StatefulInteractiveElement, Styled, StyledImage as _, Window, div, img,
     prelude::FluentBuilder as _, px, relative, rems,
 };
 use markdown::mdast;
 use ropey::Rope;
+use unicode_segmentation::UnicodeSegmentation;
 
 use crate::{
-    ActiveTheme as _, Icon, IconName, StyledExt, h_flex,
+    ActiveTheme as _, ElementExt, Icon, IconName, StyledExt, h_flex,
     highlighter::{HighlightTheme, SyntaxHighlighter},
     text::{
         CodeBlockActionsFn,
@@ -387,6 +388,20 @@ impl Paragraph {
         }
 
         text
+    }
+
+    fn plain_text(&self) -> String {
+        self.children
+            .iter()
+            .map(|child| {
+                if let Some(image) = &child.image {
+                    image.alt.clone().unwrap_or_default().to_string()
+                } else {
+                    child.text.to_string()
+                }
+            })
+            .collect::<Vec<_>>()
+            .join("")
     }
 }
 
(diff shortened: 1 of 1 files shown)


huacnlee:
It seems the implementation direction might be wrong, as it uses a lot of GPUI basic layout text APIs, which shouldn't be so complicated.

huacnlee:
Close by https://github.com/longbridge/gpui-component/pull/2221

PR #2210, opened 2026-04-01
table: Add table multi row table header support

xrtxn:
Closes #[issue number]

## Description

This makes it possible to create custom headers which can group headers. This makes the table a lot more usable for a lot of columns. 

One issue I found, which is when a column grouped across fixed and non fixed columns it doesn't display correctly, there is no other issues, and this behaviour should not happen.

## Screenshot

| image | image |

## How to Test

Run the story crate, there is an example implementation

## Checklist

### DIFF
--- a/crates/story/src/stories/data_table_story.rs
+++ b/crates/story/src/stories/data_table_story.rs
@@ -192,6 +192,7 @@ struct StockTableDelegate {
     loading: bool,
     lazy_load: bool,
     full_loading: bool,
+    show_group_headers: bool,
     clicked_row: Option<usize>,
     eof: bool,
     visible_rows: Range<usize>,
@@ -266,6 +267,7 @@ impl StockTableDelegate {
             ],
             loading: false,
             full_loading: false,
+            show_group_headers: true,
             eof: false,
             visible_cols: Range::default(),
             visible_rows: Range::default(),
@@ -339,6 +341,9 @@ impl TableDelegate for StockTableDelegate {
     }
 
     fn group_headers(&self, cx: &App) -> Option<Vec<Vec<ColumnGroup>>> {
+        if !self.show_group_headers {
+            return None;
+        }
         Some(
         vec![
             vec![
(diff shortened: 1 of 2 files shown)


crates/story/src/stories/data_table_story.rs
```diff
@@ -335,6 +338,42 @@ impl TableDelegate for StockTableDelegate {
         self.columns[col_ix].clone()
     }
 
+    fn header_rows(&self, cx: &App) -> Vec<Vec<ColumnHeader>> {
```
huacnlee: Return Option for avoid create `Vec` when we not need headers

crates/ui/src/table/state.rs
```diff
@@ -166,6 +166,16 @@ impl TableVisibleRange {
...
 /// });
 /// ```
+#[derive(Clone)]
+pub(crate) struct HeaderCell {
+    pub label: SharedString,
+    pub width: Pixels,
+    pub col_span: usize,
+    pub is_leaf: bool,
+    pub leaf_col_ix: Option<usize>,
+    pub start_leaf_col_ix: usize,
```
huacnlee: This not need pub

huacnlee:
image

Scrollbar top need fix, it should below than the table header.

crates/story/src/stories/data_table_story.rs
```diff
@@ -335,6 +338,42 @@ impl TableDelegate for StockTableDelegate {
...
+                    span: 14,
+                },
+                ColumnHeader::Group {
+                    label: "Market Data".into(),
+                    span: self.columns_count(cx) - 25,
+                },
+            ],
+            (0..self.columns_count(cx))
+                .map(|i| ColumnHeader::Leaf(self.column(i, cx)))
+                .collect::<Vec<_>>(),
```
huacnlee: This should be the internal feature. Unless that, we need this code when we need custom header groups in every time.

xrtxn: Should this be automatic? What if someone doesn't want the column names to appear only the groups?

crates/story/src/stories/data_table_story.rs
xrtxn: That is problematic to implement, because currently the table sets the cells width from columns, so this results in broken views.
image

crates/story/src/stories/data_table_story.rs
huacnlee: I think it's more appropriate not to provide an option to hide the columns header, rather than risking errors.

We've been using this in practice for a long time and rarely have this requirement. Common needs should be prioritized.

xrtxn:
How does this look to you?

huacnlee:
Thank you, this feature is useful and to implement is clean.

xrtxn:
Thank you for the review!

PR #2216, opened 2026-04-03
input: Fix Input may infinite repaint loop in occurs at fractional DPI scale factors (e.g. 125%, 150%).

lyl2dora:
Fix an infinite repaint loop (perpetual `cx.notify()`) in the Input component
  that occurs at fractional DPI scale factors (e.g. 125%, 150%).

  **Root cause:** taffy's layout rounding produces sub-pixel floating-point
  differences between frames. `paint()` unconditionally called `cx.notify()` at
  the end, and `update_scroll_offset` / `set_input_bounds` also called
  `cx.notify()` without checking whether the value actually changed — creating a
  `paint → notify → repaint → paint → …` infinite loop.

  **Fix (3 commits):**
  1. Add physical-pixel granularity guards to `update_scroll_offset` and
     `set_input_bounds` — only `cx.notify()` when the rounded physical pixel
     value actually changes.
  2. Remove the unconditional `cx.notify()` at the end of `paint()`, since the
     two functions above now notify conditionally.
  3. Cap `edge_height` and `top_bottom_margin` to viewport height, preventing
     false scroll triggers when taffy rounds `bounds.height` below `line_height`.

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -1555,6 +1555,10 @@ impl InputState {
         } else {
             line_height
         };
+        // Cap margin to viewport height: taffy edge-rounding can shrink bounds.height
+        // below line_height (e.g. 34.5 phys → 34 phys at 1.5× scale), and an
+        // uncapped margin would falsely trigger scroll on every cursor movement.
+        let edge_height = edge_height.min(bounds.size.height);
         if row_offset_y - edge_height + line_height < -scroll_offset.y {
             // Scroll up
             scroll_offset.y = -row_offset_y + edge_height - line_height;
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -1442,7 +1442,7 @@ impl InputState {
         let delta = event.delta.pixel_delta(line_height);
 
         let old_offset = self.scroll_handle.offset();
-        self.update_scroll_offset(Some(old_offset + delta), cx);
+        self.update_scroll_offset(Some(old_offset + delta), window.scale_factor(), cx);
 
         // Only stop propagation if the offset actually changed
         if self.scroll_handle.offset() != old_offset {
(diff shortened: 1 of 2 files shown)


huacnlee:
Would you please upload some screenshot for this **infinite repaint loop**?

lyl2dora:
Reproduction (Windows 11, 150% DPI scaling)

  Modified examples/input to simulate inline rename — a common pattern for file trees / tables:

  Input::new(&self.input_state)
      .appearance(false)
      .px_0()
      .py_0()
      .h_full()

https://github.com/user-attachments/assets/fce39f11-1709-4072-9332-58cb6c10b537

Each click shifts the text down by ~0.33px (0.5 physical pixel at 1.5×). The root cause:

  1. taffy rounds a fractional physical height (e.g. n.5 → n), making bounds.height slightly less than line_height
  2. scroll_to() uses line_height as edge margin — when margin > viewport, it misjudges the cursor as out of bounds and
  applies a negative scroll offset
  3. paint() unconditionally calls cx.notify(), and update_scroll_offset / set_input_bounds compare floats with !=, so
  sub-pixel noise keeps triggering repaints indefinitely

  This affects any Input used for inline editing (zero padding, tight height) at fractional DPI scales (125%, 150%).

crates/ui/src/input/state.rs
```diff
@@ -1476,8 +1477,16 @@ impl InputState {
...
         offset.x = offset.x.clamp(safe_x_range.start, safe_x_range.end);
+
+        // Compare at physical-pixel granularity to avoid sub-pixel oscillation
+        // that would cause a perpetual notify→repaint loop.
+        let old = self.scroll_handle.offset();
         self.scroll_handle.set_offset(offset);
-        cx.notify();
+
+        let snap = |v: Pixels| (v.as_f32() * scale_factor).round() as i32;
+        if snap(old.x) != snap(offset.x) || snap(old.y) != snap(offset.y) {
```
huacnlee, line 1487: This conversion process is strange, and I don't think it's a solution to the problem.

huacnlee (changes requested):

huacnlee:
Close too long no response PRs.

PR #2230, opened 2026-04-07
table: Add `set_right_clicked_row` method.

### LINKED ISSUE #2214: table: no way to prevent row context menu when right-clicking header
## Problem

When using `TableDelegate::render_th()` with `ContextMenuExt` to add context menus to column headers, right-clicking a header shows **both** the row context menu and the header context menu simultaneously.

### Root cause

The row context menu is attached to `inner_table` (in `render()` at state.rs ~line 1973), which wraps both the header and body. When right-clicking a header cell:

1. The `inner_table` context menu fires because the header is geometrically inside its bounds
2. It reads `right_clicked_row` which may still be `Some` from a previous row click
3. The header's own context menu (from `ContextMenuExt`) also opens independently
4. Both menus appear at the same time

The `on_mouse_down_out` dismiss handler on the row menu doesn't fire because the click is **inside** `inner_table`'s bounds.

### Workaround

I overrode `render_header` to add `on_mouse_down(MouseButton::Right)` that clears the row state before menu builders run. However, `right_clicked_row` is private, so I forked and added:

```rust
pub fn clear_right_clicked_row(&mut self) {
    self.right_clicked_row = None;
}
```

### Suggested fix

Either:
1. **Expose `clear_right_clicked_row()`** (or a `set_right_clicked_row`) as a public method on `TableState`
2. **Add a `header_context_menu()` method** to `TableDelegate` (similar to `context_menu()` for rows) that the framework handles correctly
3. **Window-level context menu tracking** so opening any context menu auto-dismisses others

Option 1 is the smallest change. Option 2 is the cleanest API.

### Reproduction

```rust
impl TableDelegate for MyDelegate {
    fn render_th(&mut self, col_ix: usize, _window: &mut Window, cx: &mut Context<TableState<Self>>) -> impl IntoElement {
        div()
            .id(("th", col_ix))
            .child("Header")
            .context_menu(|menu, _w, _cx| {
                menu.item(PopupMenuItem::new("Header action"))
            })
    }

    fn context_menu(&mut self, row_ix: usize, menu: PopupMenu, ...) -> PopupMenu {
        menu.item(PopupMenuItem::new("Row action"))
    }
}
```

Steps: Right-click a row → right-click a column header → two menus appear.

boboshan:
Fixes #2214

## Summary

- Adds `clear_right_clicked_row()` public method to `TableState`
- Allows delegates to clear the row right-click state from `render_header`, preventing row and header context menus from appearing simultaneously

## AI Disclosure

This code was generated with AI assistance (Claude). The implementation has been manually tested in a production GPUI application.

### DIFF
--- a/crates/ui/src/table/state.rs
+++ b/crates/ui/src/table/state.rs
@@ -392,6 +392,15 @@ where
         self.right_clicked_row
     }
 
+    /// Set or clear the right-clicked row state.
+    ///
+    /// Pass `None` to clear — useful when opening a header context menu
+    /// to prevent the row context menu from appearing simultaneously.
+    pub fn set_right_clicked_row(&mut self, row: Option<usize>, cx: &mut Context<Self>) {
+        self.right_clicked_row = row;
+        cx.notify();
+    }
+
     /// Returns the selected column index.
     pub fn selected_col(&self) -> Option<usize> {
         self.selected_col


crates/ui/src/table/state.rs
```diff
@@ -392,6 +392,14 @@ where
         self.right_clicked_row
     }
 
+    /// Clear the right-clicked row state.
+    ///
+    /// Useful when opening a header context menu to prevent the row
+    /// context menu from appearing simultaneously.
+    pub fn clear_right_clicked_row(&mut self) {
```
huacnlee: Rename to set_right_clicked_row, this is acceptable.

huacnlee (changes requested):

crates/ui/src/table/state.rs
boboshan: Done — renamed to `set_right_clicked_row(row: Option<usize>)`. Force-pushed.

crates/ui/src/table/state.rs
```diff
@@ -392,6 +392,14 @@ where
         self.right_clicked_row
     }
 
+    /// Set or clear the right-clicked row state.
+    ///
+    /// Pass `None` to clear — useful when opening a header context menu
+    /// to prevent the row context menu from appearing simultaneously.
+    pub fn set_right_clicked_row(&mut self, row: Option<usize>) {
+        self.right_clicked_row = row;
+    }
```
huacnlee, line 402: `pub fn set_selected_col(&mut self, col_ix: usize, cx: &mut Context<Self>)`

Set method should have `cx` argument, and call `cx.notify()`

boboshan: Done — added `cx: &mut Context<Self>` parameter and `cx.notify()` call. Force-pushed.

huacnlee (approved):

PR #2235, opened 2026-04-09
input: Allow disabling the mouse context menu

jstnd:
## Description

These changes allow for the use case of building a custom context menu within the input component without showing the built-in context menu. There appears to be no other way to disable the built-in mouse context menu before these changes (as far as I could see, please correct if wrong).

**Sidenote:** 
My use case is wanting to add my own options inside this context menu, but there doesn't seem to be a way to customize the built-in menu. So, I went with just making my own context menu entirely using `context_menu` on a parent element, but the built-in context menu still shows underneath.

Please let me know if there's a more ideal way of integrating this use case.

## Checklist


madcodelife:
Thank you, we'll consider adding support for customizing the context menu.

PR #2237, opened 2026-04-10
select: Focus Select when closing

dunkmann00:
## Description

When closing the select component, make it take focus back from the list. This helps when used in a dialog to not block other mouse interactions. This also helps with keyboard navigation.

The defer is necessary to avoid the list recapturing focus if the select box is clicked on to close.

## Screenshot

| <video src="https://github.com/user-attachments/assets/d55cda72-9ee7-423f-9634-a8d22a2302f1"> | <video src="https://github.com/user-attachments/assets/99752b62-aba9-43d9-9967-a06bf11b29dc"> |

## How to Test

Use the gallery dialog example or the select example.

## Checklist


huacnlee (dismissed):
This behavior actually needs to improve, but:

We were tried for this. Use defer will get unexpected behavior.

For example if some synchronized calls  focus directly, this defer will catch focus back again.

dunkmann00:
Okay, I'm not 100% sure I understand when to use vs not to use defer so I figured I would try it.

I've updated it with a different approach that seems like it might work. But this uses `cx.stop_propagation()`, so I'm also not sure if this is appropriate to use here or if it to might cause problems.

I did notice that when you close the select by pressing the `ESC` key there are no problems, and there were also no problems when you click on an item in the list. In both of those cases it closes the list and gives focus back to the select state. I checked out the code that handles that in `SelectListDelegate` and it seems that `defer` is used there.

Cancel:
https://github.com/longbridge/gpui-component/blob/2f7c7445d8c55e4af95c2a22695184ea8df35a32/crates/ui/src/select.rs#L240

Confirm:
https://github.com/longbridge/gpui-component/blob/2f7c7445d8c55e4af95c2a22695184ea8df35a32/crates/ui/src/select.rs#L258

Does the defer not cause any issue when used in this way?

madcodelife:
Thank you. I’ve made a small change, added an early return.

dunkmann00:
Great, thank you!

PR #2265, opened 2026-04-18
input: fix multiline placeholder shaping in empty multiline inputs

### LINKED ISSUE #2264: input: multiline placeholder shaping uses mismatched runs in empty multiline inputs
## Summary

When an input is empty and its placeholder contains multiple lines, the empty-placeholder layout path can shape each placeholder line with `TextRun`s that were built for the whole placeholder string.

That breaks the same invariant used by the normal multiline layout path: the text passed to `shape_line` must match the supplied runs.

## Affected conditions

- multiline input mode (`!state.mode.is_single_line()`)
- empty text (`state.text.len() == 0`)
- placeholder contains `\n`

## Why this is a bug

In `TextElement::layout_lines`, the normal multiline path slices runs per wrapped sub-line with `runs_for_range(...)` before calling `shape_line(...)`.

The empty-placeholder branch used to do something different:

1. build `display_text` from the full placeholder
2. build `runs` whose total length matches the full placeholder
3. split the placeholder by `\n`
4. call `shape_line(line, ..., &runs, ...)` for each split line

For a placeholder like `"a\nb"`, the first shaping call receives:

- text: `"a"`
- runs: lengths that describe `"a\nb"`

So the shaping input is internally inconsistent.

## Impact

This is not just a cosmetic placeholder issue.

Depending on the text backend / platform, this can fail during shaping and may panic or crash during the initial render of an empty multiline input. Historically this kind of mismatch is most likely to surface as an explicit failure on Windows, but the call pattern is invalid regardless of platform.

Affected surfaces include:

- multiline `Input`
- textarea-style inputs
- editor-like inputs
- any future empty-text + multiline-placeholder case

## Reproduction

Use an empty multiline input with a placeholder that contains a newline, for example:

```rust
placeholder = "a\nb"
text = ""
mode = multiline

### PR DESCRIPTION
**Pull request by lurenjia534**
Closes #2264

## Description

This PR fixes multiline placeholder shaping for empty multiline inputs.

Previously, when an input was empty and its placeholder contained `\n`, the empty-placeholder layout path split the placeholder into lines but reused `TextRun`s built for the entire placeholder string for each `shape_line(...)` call. That could make the shaping input internally inconsistent and cause empty multiline inputs to fail during initial render.

This change keeps the fix scoped to this single issue:

- slice placeholder runs per visual line before calling `shape_line(...)`
- keep multiline placeholder lines inside a single `LineLayout` so layout metadata stays aligned with the empty buffer
- handle leading empty visual lines correctly when resolving cursor / IME positions

This fixes empty multiline inputs such as textarea- or editor-like fields with multiline placeholders and avoids placeholder-specific layout inconsistencies.

## Screenshot

| Before | After |
| --- | --- |
| Empty multiline input with a multiline placeholder could fail during initial render or show incorrect placeholder layout. | Multiline placeholder lines render correctly in empty multiline inputs, and cursor / IME positioning stays aligned. |

## How to Test

1. Run the focused unit tests:

```bash
cargo test -p gpui-component test_placeholder_line_runs
cargo test -p gpui-component test_position_for_index_prefers_first_leading_empty_visual_line
```
2. Run the story app:

3. Open a multiline input / textarea-style story and use a placeholder such as ab\n\nc.

4. Leave the input empty and confirm:

   * it renders without crashing
   * each placeholder line is displayed correctly
   * leading empty placeholder lines do not move the cursor / IME position to a later visual line
   
   ## Checklist

* [x] Reviewed the changes in this PR and confirmed AI generated code (If any) is accurate.
* [x] Passed `cargo run` for story tests related to the changes.
* [ ] Tested macOS, Windows and Linux platforms performance (if the change is platform-specific)

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -1925,6 +1925,9 @@ fn placeholder_line_runs<'a>(
     result
 }
 
+/// Get the runs for the given range.
+///
+/// The range is the byte range of the wrapped line.
 pub(super) fn runs_for_range(
     runs: &[TextRun],
     line_offset: usize,
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -961,21 +961,24 @@ impl TextElement {
 
         // Empty to use placeholder, the placeholder is not in the wrapper map.
         if state.text.len() == 0 {
-            return display_text
-                .to_string()
-                .split("\n")
-                .map(|line| {
-                    let shaped_line = window.text_system().shape_line(
-                        line.to_string().into(),
-                        font_size,
-                        &runs,
-                        None,
-                    );
-                    LineLayout::new()
-                        .lines(smallvec::smallvec![shaped_line])
-                        .with_whitespaces(whitespace_indicators.clone())
-                })
-                .collect();
+            let placeholder_text = display_text.to_string();
+            let mut placeholder_lines = SmallVec::new();
+
+            for (line, line_runs) in placeholder_line_runs(&placeholder_text, runs) {
+                let shaped_line = window.text_system().shape_line(
+                    line.to_string().into(),
+                    font_size,
+                    &line_runs,
+                    None,
+                );
+                placeholder_lines.push(shaped_line);
+            }
+
+            // Keep placeholder lines in a single layout to stay parallel with visible_* metadata.
+            let line_layout = LineLayout::new()
+                .lines(placeholder_lines)
+                .with_whitespaces(whitespace_indicators);
+            return vec![line_layout];
         }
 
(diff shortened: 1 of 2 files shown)


crates/ui/src/input/element.rs
```diff
@@ -1900,9 +1903,28 @@ impl Element for TextElement {
     }
 }
 
-/// Get the runs for the given range.
```
madcodelife, line 1903: This comment seems to have been deleted by mistake.

lurenjia534: I restored the runs_for_range doc comment in f24fbcea. Thank you for the review !

madcodelife:
Thank you, nice catch!

PR #2267, opened 2026-04-19
input: Improve InputState set_value behavior

### LINKED ISSUE #2262: input: Do not let `set_value` trigger the input change event
When listening to input change, setting the value will be repeated and is hard to handle.

dunkmann00:
Closes #2262

## Description

This adds a flag to suppress events. This is used to prevent emiting an `InputEvent::Change` when setting the text programatically with set_value.

This also only pauses blinking if the cursor was already visible. Otherwise this can cause unnecessary rendering.

This also clears the history after changing the value this way. I don't think the history is correct after having changed it this way so it makes more sense to reset it.

## Checklist

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -372,6 +372,8 @@ pub struct InputState {
     _pending_update: bool,
     /// A flag to indicate if we should ignore the next completion event.
     pub(super) silent_replace_text: bool,
+    /// A flag to indicate we should not emit InputEvents.
+    pub(super) suppress_events: bool,
 
     /// To remember the horizontal column (x-coordinate) of the cursor position for keep column for move up/down.
     ///
@@ -460,6 +462,7 @@ impl InputState {
             hover_popover: None,
             hover_definition: HoverDefinition::default(),
             silent_replace_text: false,
+            suppress_events: false,
             size: Size::default(),
             _subscriptions,
             _context_menu_task: Task::ready(Ok(())),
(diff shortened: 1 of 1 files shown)


crates/ui/src/input/state.rs
```diff
@@ -697,8 +700,10 @@ impl InputState {
         cx: &mut Context<Self>,
     ) {
         self.history.ignore = true;
+        self.suppress_events = true;
         self.replace_text(value, window, cx);
         self.history.ignore = false;
+        self.suppress_events = false;
```
huacnlee: Not only `set_value`, for example when `paste` or `cut`, we should emit events.

huacnlee: Your changes has been changed default behaviour of the emit `Input::Change` event.

We should not able to do like this, we need keep the original behaviour.

dunkmann00: This change only effects `set_value`? Paste or cut should not be affected by this unless I am misunderstanding how this works? The flag is just switched on for the `replace_text` call in `set_value`, then turned back off to let the change event happen like before.

dunkmann00: I thought based on #2262, we don't want to emit an event when setting the value programmatically? I think this makes sense to not emit in this case, since we are doing it, not the user. Also as is mentioned in that issue, it makes handling changes more difficult since you might create a loop where you `set_value` then get a subscribe handler call, and then keep repeating that loop.

huacnlee: Oh, I know the design. Yes , the behaviour are same like before.

Now, please renamed this `suppress_events` to `emit_events` and set default to `true`.

dunkmann00:
Name has been updated to `emit_events`!

PR #2274, opened 2026-04-21
chore: Adapt to latest GPUI API changes (Corner → Anchor, Corners API)

elcoosp:
Remove the custom `anchored.rs` fork that was added when GPUI lacked this functionality. GPUI now provides an equivalent built-in `Anchored` element, so migrate all usages to `gpui::anchored` directly.

- Delete `crates/ui/src/anchored.rs`
- Update `tooltip.rs` and `popover.rs` to import `anchored` from `gpui`
- Replace `position_fn()` (not available in GPUI's API) with `position()` in `popover.rs`


huacnlee:
Thanks, I am also want to do this.

PR #2277, opened 2026-04-21
date_picker: Add support to set year range

### LINKED ISSUE #2276: Configuring the years range on a DatePicker
I was making a `DatePicker` for a date that could be more than 50 years in the past (i.e. birthday). The current `DatePicker` implementation only shows years going back 50 years ago (when in years mode). It seems the calendar years range can be configured on `CalendarState`, but with `DatePicker`, it cannot.

I was wondering if this functionality is something you would consider adding/accepting a PR for?

dunkmann00:
Closes #2276 

## Description

This adds a way to make a DatePickerState with a provided CalendarState. This enables the option to configure the calendar state that is used in the date picker. One way this can be useful is with setting a custom years range.

This also adds this feature as an example on the date picker story.

I figured I would put this together in case this approach is acceptable to you. It was the simplest way I could think of to accomplish setting the years range on the date picker. Totally understand if you don't like exposing the `CalendarState` like this.

## How to Test

I added an example in the date picker story under the new section "Custom Calendar Year Range". When viewing the calendar in years mode it ranges from this year and goes back 99 years (a 100 year range).

## Checklist


madcodelife:
Thank you, I tried to add support with minimal changes, which also better aligns with our project’s conventions.

dunkmann00:
Works for me, thank you!

PR #2278, opened 2026-04-22
input: Expose selected_range as a public getter

glani:
## Motivation

`InputState::selected_range` is currently `pub(super)`, reachable only
inside the `input` module. Downstream code that wraps `Input` — custom
editors, diff/merge surfaces, gutter overlays that highlight the
selected region — needs to observe the current selection to emit
selection-change events, drive decorations, or feed search-and-replace
UIs without reimplementing selection tracking on top of the widget.

## API shape

Adds one read-only getter:

```rust
impl InputState {
    pub fn selected_range(&self) -> std::ops::Range<usize>;
}
```

The return value is an empty range (`start == end`) when nothing is
selected; in that case the offset equals `cursor()`. Byte offsets are
measured in the underlying rope's byte units, matching the convention
used elsewhere in the module.

## Behavior

No behavior change. The implementation is a one-line accessor that
converts the existing internal `Selection` via its `Into<Range<usize>>`
impl. No existing call sites are touched.


madcodelife:
Thank you.

PR #2279, opened 2026-04-22
input: Expose viewport accessors on InputState

glani:
## Motivation

`InputState` already tracks the laid-out viewport (visible row range,
line height) and owns the scroll handle, but none of it is reachable
outside the `input` module. Downstream code that renders aligned
gutters, sticky minimap overlays, scroll-synced decorations, or
viewport-scoped syntax highlighting has to duplicate this layout state
or reach through private fields. Exposing what's already computed
removes that duplication.

## API shape

Adds three read-only getters on `InputState`:

```rust
impl InputState {
    pub fn visible_row_range(&self) -> Option<std::ops::Range<usize>>;
    pub fn scroll_offset(&self)     -> gpui::Point<gpui::Pixels>;
    pub fn line_height(&self)       -> Option<gpui::Pixels>;
}
```

`visible_row_range` and `line_height` return `None` before the first
layout pass, since `last_layout` isn't populated yet. `scroll_offset`
is always available from the scroll handle.

## Behavior

No behavior change. Each getter is a one-liner that reads data the
struct already stores. No existing call sites are touched.


madcodelife:
I’ve resolved the conflict, Thank you.

PR #2286, opened 2026-04-24
sidebar: Add click_to_toggle functionality to SidebarMenuItem

BeratHundurel:
## Description

This adds click_to_toggle to SidebarMenuItem, letting submenu items toggle their open state when the user clicks the item itself.

Previously, submenu open state was internal and not writable from user code, so implementing toggle behavior externally was not possible (at least I was not able to do it). With this change, the sidebar can support user-facing toggle interactions without exposing internal state.

click_to_open still takes precedence, and the existing caret-only behavior remains unchanged unless toggle behavior is enabled.

I wasn’t entirely sure whether exposing open_state is the right approach. If click_to_toggle isn’t aligned with the preferred direction, I’m happy to open another PR following a different pattern. Also, if there’s already a way to achieve this behavior without introducing this change, I’d really appreciate any guidance.

Breaking Changes
None.

## Checklist

### DIFF
--- a/crates/ui/src/sidebar/menu.rs
+++ b/crates/ui/src/sidebar/menu.rs
@@ -97,6 +97,7 @@ pub struct SidebarMenuItem {
     default_open: bool,
     click_to_open: bool,
     collapsed: bool,
+    click_to_toggle: bool,
     children: Vec<Self>,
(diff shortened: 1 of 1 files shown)


madcodelife:
Thank you.

PR #2292, opened 2026-04-25
resizable: Support style overrides on ResizablePanel

boboshan:
## Description

`ResizablePanel` does not currently implement `Styled`, so call sites
cannot override the panel's rendered styles. The load-bearing case is
flex behavior: the panel sets `flex_grow: 1` unconditionally in its
render, so a sized panel inherits growth and absorbs slack from
siblings — most visibly when a sibling toggles its `.visible()`.

This PR adopts the same pattern `Button` and `Input` use:

- `style: StyleRefinement` field on `ResizablePanel`
- `impl Styled for ResizablePanel`
- `.refine_style(&self.style)` applied last in render so call-site
  overrides win over internal flex defaults

The internal `flex_grow: 1` is unchanged — there is **no behavior
change for existing call sites**. To opt out of growth, write
`.flex_none()` in the chain. It overrides `flex_grow` / `flex_shrink`
while leaving `flex_basis` (managed by `ResizableState`) untouched.

The struct doc enumerates reserved styles to avoid (`flex_basis`,
`absolute`, `overflow_hidden`).

## Screenshot

The Resizable story gains a third demo with two buttons:
**Hide Left** (toggles a sibling's `.visible()`) and **Use flex_none**
(toggles the override on the same layout). Recording shows the same
layout under both states.

https://github.com/user-attachments/assets/eb8b6662-4989-4917-95ca-4dc5572681cb

## Break Changes

None. The default `StyleRefinement::default()` is empty, so render
output is byte-identical for any caller that does not invoke `Styled`
methods.

## How to Test

Open the **Resizable** story. The new demo at the bottom has two
buttons next to a 3-panel `h_resizable`:

1. With **Use flex_none: ON** → click **Hide Left**. The right panel
   should hold its width; the center flex panel absorbs the freed
   space.
2. Click **Show Left** to reset, then click **Use flex_none: OFF** →
   click **Hide Left** again. The right panel grows by half of the
   freed space.

## Checklist

- [x] Reviewed the changes in this PR. AI assistance noted in the
      commit body and below; the foot-gun list and story example
      were authored deliberately, and the whole branch was reviewed
      against the story before commit.
- [x] Passed the Resizable story for the new demo.

## AI assistance

Iterated with Claude Code: implementation, story example, doc comment.
Each step reviewed and tested locally.

### DIFF
--- a/crates/ui/src/resizable/panel.rs
+++ b/crates/ui/src/resizable/panel.rs
@@ -6,10 +6,11 @@ use std::{
 use gpui::{
     Along, AnyElement, App, AppContext, Axis, Bounds, Context, Element, ElementId, Empty, Entity,
     EventEmitter, InteractiveElement as _, IntoElement, IsZero as _, MouseMoveEvent, MouseUpEvent,
-    ParentElement, Pixels, Render, RenderOnce, Style, Styled, Window, div, prelude::FluentBuilder,
+    ParentElement, Pixels, Render, RenderOnce, Style, StyleRefinement, Styled, Window, div,
+    prelude::FluentBuilder,
 };
 
-use crate::{AxisExt, ElementExt, h_flex, resizable::PANEL_MIN_SIZE, v_flex};
+use crate::{AxisExt, ElementExt, h_flex, resizable::PANEL_MIN_SIZE, styled::StyledExt as _, v_flex};
 
 use super::{ResizableState, resizable_panel, resize_handle};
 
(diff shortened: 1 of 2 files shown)


crates/ui/src/resizable/panel.rs
huacnlee, line 290: Move `.refine_style(&self.style)` to here, unless that, it will override the width/height to get unexpected result.

boboshan:
Moved `.refine_style(&self.style)` from after the size constraints to right after `.relative()` (before `min_w/max_w/min_h/max_h` and the `flex_basis(panel_state.size)` chain). The panel's runtime size management is now authoritative and caller styles can't accidentally clobber width/height.

The original load-bearing case still works: `.flex_grow()` is set above the refine call, so a caller's `.flex_none()` cancels it. Padding / colors / borders also still take effect (they're not in the size-management chain).

Updated the struct-field doc and the inline render comment to match. Force-pushed.

PR #2293, opened 2026-04-25
setting: Make dropdown popup scrollable by default with fluent override

boboshan:
## Summary

`SettingField`'s dropdown variant builds its popup via `PopupMenu` but never opts into `PopupMenu::scrollable`. With long option lists the menu overflows the viewport and entries below the fold are unreachable.

This PR:

- Defaults `scrollable: true` so out-of-the-box dropdowns degrade gracefully on long menus.
- Adds a fluent `.scrollable(bool)` setter on `SettingField<SharedString>` for callers who want to opt out, mirroring the builder style of `GroupBoxVariants::with_variant` and `Sizable::small`/`medium`.

```rust
SettingField::dropdown(items, getter, setter)              // scrollable
SettingField::dropdown(items, getter, setter)
    .scrollable(false)                                      // no scroll
```

`SettingFieldType::Dropdown` gains a `scrollable: bool` field; the existing `dropdown(...)` constructor sets it to `true`. The setter mutates the variant in place when the field is a dropdown, no-op otherwise.

No public API breakage — all existing callers of `SettingField::dropdown(...)` keep working unchanged.

## Test plan

- [x] `cargo check --workspace` passes locally.
- [x] Long option list (verified in a downstream app): menu scrolls, items below fold reachable.
- [x] Short option list still renders without scrollbar chrome.
- [ ] `dropdown(...).scrollable(false)` disables scrolling.

### DIFF
--- a/crates/ui/src/setting/fields/mod.rs
+++ b/crates/ui/src/setting/fields/mod.rs
@@ -60,6 +60,7 @@ pub enum SettingFieldType {
     Input,
     Dropdown {
         options: Vec<(SharedString, SharedString)>,
+        scrollable: bool,
     },
     Element {
         element: Rc<dyn SettingFieldElement<Element = AnyElement>>,
@@ -95,8 +96,16 @@ impl SettingFieldType {
     #[inline]
     pub(super) fn dropdown_options(&self) -> Option<&Vec<(SharedString, SharedString)>> {
         match self {
-            SettingFieldType::Dropdown { options } => Some(options),
+            SettingFieldType::Dropdown { options, .. } => Some(options),
             _ => None,
+        }
+    }
+
+    #[inline]
+    pub(super) fn dropdown_scrollable(&self) -> bool {
+        match self {
+            SettingFieldType::Dropdown { scrollable, .. } => *scrollable,
+            _ => false,
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/ui/src/setting/fields/mod.rs
```diff
@@ -168,7 +180,25 @@ impl SettingField<SharedString> {
...
+            value,
+            set_value,
+        )
+    }
+
+    /// Set whether the dropdown popup menu scrolls when its content exceeds
+    /// the viewport, default is `true`.
+    ///
+    /// Has no effect on non-dropdown field types.
+    pub fn scrollable(mut self, scrollable: bool) -> Self {
```
madcodelife: Thanks for addressing the overflow issue.

The `scrollable(bool)` method is somewhat opaque — it would be better to just add a dedicated `scrollable_dropdown()` method to distinguish it from the existing `dropdown()`.

```rust
SettingField::dropdown(items, getter, setter)            // unchanged, no scroll
SettingField::scrollable_dropdown(items, getter, setter) // opt-in scroll
```

boboshan:
Rewrote per the suggestion — replaced the fluent `.scrollable(bool)` setter with a sibling constructor:

```rust
SettingField::dropdown(items, ...)            // unchanged, no scroll
SettingField::scrollable_dropdown(items, ...) // opt-in scroll
```

`dropdown(...)` now keeps `scrollable: false`, so existing call sites are byte-identical. The new constructor is purely additive. Tested locally — short dropdowns stay non-scrolling, long ones (21-item theme picker) scroll cleanly. Force-pushed.

madcodelife:
Thank you!

PR #2294, opened 2026-04-25
resizable: Expose `resize_panel` method for programmatic resize

boboshan:
## Summary

`ResizableState` exposes only the `sizes()` getter — `self.sizes` can be mutated solely by user-driven handle drags. The two existing patterns documented for layout work fine on their own:

- **Initial layout from preferences** — `ResizablePanel::size(...)` is honored on first render (while the slot is still at `PANEL_MIN_SIZE`).
- **Persisting user drags** — subscribe to `ResizablePanelEvent::Resized`.

What's missing is **imperative resize after init**:

- "Compact / expand" toggle that snaps a panel to a fixed size on click.
- "Reset to default layout" action.
- Syncing a panel's size to non-drag input (a number input, an action, another panel's state).

All of these currently require forking.

### Change

- New public `pub fn resize_panel(&mut self, ix, size, window, cx)`.
  Out-of-range indices no-op. Non-last panels go through the existing handle-walk redistribution; the last panel is driven by resizing its predecessor so the freed space lands on it. Emits the same `ResizablePanelEvent::Resized` as a drag-finish so subscribers see the change uniformly.
- The previous private `fn resize_panel(...)` is renamed `fn resize_panel_at_handle(...)` since its `ix` parameter is the drag-handle index, not the panel index. The two internal call sites in `panel.rs` are updated.
- `ResizableStory` gains a fourth demo block: four buttons (*Compact left → 100*, *Reset left → 200*, *Compact right → 80*, *Reset right → 200*) drive a 3-panel `h_resizable` via `ResizableState::resize_panel`, exercising both the non-last and last-panel paths.

The public name matches the verb-first convention of the rest of `ResizableState`'s mutators (`insert_panel`, `remove_panel`, `replace_panel`).

## Test plan

- [x] `cargo check -p gpui-component-story` passes locally.
- [x] Drag-resize behavior in the resizable story is unchanged (the rename is the only thing that touches the drag path).
- [x] Calling `state.resize_panel(ix, size, window, cx)` from a host app moves the panel to `size`, redistributes among siblings, and fires `ResizablePanelEvent::Resized`. Verified end-to-end against a "compact panel" toggle in a downstream app — round-trip 250 → 60 → 250 produced clean `[layout, waveform]` sizes at every step.
- [x] Last-panel resize drives the previous sibling rather than silently doing nothing.

### DIFF
--- a/crates/story/src/stories/resizable_story.rs
+++ b/crates/story/src/stories/resizable_story.rs
@@ -3,10 +3,10 @@ use gpui::{
     ParentElement as _, Pixels, Render, SharedString, Styled, Window, div, px,
 };
 use gpui_component::{
-    ActiveTheme,
+    ActiveTheme, Sizable as _,
     button::Button,
     h_flex,
-    resizable::{h_resizable, resizable_panel, v_resizable},
+    resizable::{ResizableState, h_resizable, resizable_panel, v_resizable},
     v_flex,
 };
 
@@ -14,6 +14,7 @@ pub struct ResizableStory {
     focus_handle: FocusHandle,
     show_left: bool,
     use_flex_none: bool,
+    programmatic_state: Entity<ResizableState>,
 }
 
 impl super::Story for ResizableStory {
@@ -46,6 +47,7 @@ impl ResizableStory {
             focus_handle: cx.focus_handle(),
             show_left: true,
             use_flex_none: true,
+            programmatic_state: cx.new(|_| ResizableState::default()),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


huacnlee:
Please add example for call this method in ResizableStory for test.

boboshan:
Added a third demo to `ResizableStory` that exercises `resize_panel` directly. The story now holds an `Entity<ResizableState>` and four buttons drive it — *Compact left → 100*, *Reset left → 200*, *Compact right → 80*, *Reset right → 200* — against a 3-panel `h_resizable` (200 / grow / 200). Both the non-last and last-panel paths are covered.

huacnlee:
This branch has conflicts

boboshan:
Rebased onto current `origin/main` (which now includes #2292's `flex_none` story demo) and resolved the `resizable_story.rs` conflict by keeping both demo blocks: the `flex_none` toggle for #2292 stays as the third panel, and the `resize_panel` 4-button demo for this PR sits as the fourth. `cargo check -p gpui-component-story` passes against the rebased state. Force-pushed.

PR #2306, opened 2026-04-27
input: Fix input auto grow scroll transition

### LINKED ISSUE #2297: Input auto_grow flickers when long text reaches max rows and internal scrolling starts
## Description

`InputState::auto_grow(min_rows, max_rows)` can visually flicker when a long text input reaches `max_rows` and the component switches from growing height to internal scrolling.

This is visible in a chat-composer style UI that renders an appearance-less `Input` inside a rounded composer container:

```rust
let input_state = cx.new(|cx| {
    InputState::new(window, cx)
        .multi_line(true)
        .auto_grow(3, 8)
        .placeholder("Ask anything")
});

div()
    .w_full()
    .rounded(px(18.))
    .p_2()
    .child(
        div()
            .flex()
            .w_full()
            .child(Input::new(&input_state).flex_1().appearance(false)),
    )
```

When pasting or typing enough wrapped text to exceed 8 rows, the input briefly renders in an intermediate state before settling into the final scrollable layout. In practice this looks like a one-frame flash/flicker in the composer.

## Environment

- GPUI: v0.2.2
- GPUI Component: current `main` / git dependency around `9e252d8667e18039c4c3a64cde02758d6cfdf47a`
- Platform: macOS 26

## Steps to Reproduce

1. Render an `InputState` with `.auto_grow(3, 8)` and `Input::new(&state).appearance(false)` inside a fixed-width container.
2. Focus the input.
3. Paste a long paragraph that wraps beyond 8 rows.
4. Observe the composer/input area while the content transitions from auto-growing to internally scrollable.

Minimal standalone example:

```rust
use gpui::{
    App, AppContext, Context, Entity, IntoElement, ParentElement, Render, Styled, Window,
    WindowOptions, div, px,
};
use gpui_component::{
    ActiveTheme as _, Sizable as _,
    input::{Input, InputState},
};

struct AutoGrowFlickerRepro {
    input: Entity<InputState>,
}

impl AutoGrowFlickerRepro {
        let input = cx.new(|cx| {
            InputState::new(window, cx)
                .multi_line(true)
                .auto_grow(3, 8)
                .placeholder("Paste long text here")
        });

        input.update(cx, |input, cx| input.focus(window, cx));

        Self { input }
    }
}

impl Render for AutoGrowFlickerRepro {
    fn render(&mut self, _window: &mut Window, cx: &mut Context<Self>) -> impl IntoElement {
        div()
            .size_full()
            .p(px(24.))
            .bg(cx.theme().background)
            .child(
                div()
                    .w(px(720.))
                    .rounded(px(18.))
                    .border_1()
                    .border_color(cx.theme().border)
                    .p(px(8.))
                    .child(
                        div()
                            .flex()
                            .w_full()
                            .child(Input::new(&self.input).flex_1().appearance(false)),
                    ),
            )
    }
}

fn main() {
    App::new().run(|cx| {
        gpui_component::init(cx);

            cx.new(|cx| AutoGrowFlickerRepro::new(window, cx))
        })
        .unwrap();
    });
}
```

Paste text such as:

```text
3. **Capture a specific window**:
- After pressing `Command + Shift + 4`, press Space. The cursor changes to a camera icon. You can click the window to capture it.

4. **Capture the Touch Bar**:
- Use `Command + Shift + 6`.

After capturing, screenshots are saved to the Desktop by default. You can also use the Screenshot app to choose save locations and options.
```

## Root Cause

The flicker appears to come from an inconsistent frame during the auto-grow-to-scroll transition.

Relevant code path:

- `InputMode::update_auto_grow` updates rows from `display_map.wrap_row_count()`.
- `InputMode::set_rows` clamps auto-grow rows to `min_rows..max_rows`.
- `TextElement::request_layout` uses `rows * line_height` as the auto-grow min height.
- `Input::render_editor` builds the scrollbar from `state.last_layout` and `state.scroll_size`.
- `TextElement::paint` writes the new `last_layout`, `scroll_size`, and scroll offset back into `InputState`, then calls `cx.notify()`.

This means the frame that first crosses `max_rows` can render the scrollbar/editor using stale layout and scroll-size data, then immediately schedule another frame with the corrected state. That one-frame mismatch is visible as a flicker.

## Expected

When pasted/typed content exceeds `max_rows`, the input should transition directly to the final scrollable state without a visible intermediate flash.

## Actual

The input briefly flashes/flickers around the moment it reaches max rows and internal scrolling begins.

## Suggested Fix

Avoid constructing the editor scrollbar from stale `InputState` layout data during auto-grow transitions.

Possible approaches:

1. Use the current prepaint/layout result to drive scrollbar size and visibility in the same frame instead of reading `state.scroll_size` from the previous paint.
2. Update auto-grow rows, input bounds, scroll size, and scroll offset as one coherent layout state before rendering the scrollbar.
3. Suppress the extra `cx.notify()` when the only change is a paint-time writeback that does not affect the already requested layout, or restructure the writeback so the next frame does not show an inconsistent intermediate state.

A regression test or story scenario should cover pasting wrapped text that crosses `max_rows` for `InputState::auto_grow(3, 8)`.

suxiaoshao:
## Summary

Fixes #2297.

Fixes the `InputState::auto_grow(3, 8)` transition when wrapped text grows past `max_rows` and the input switches to internal scrolling.

The flicker came from one-frame mismatches between current prepaint layout data and state from the previous paint. This patch builds editor scrollbar geometry from the current frame and clamps auto-grow scroll offsets against the current viewport before text and scrollbar painting.

## Changes

- Build the editor scrollbar in `TextElement::prepaint` from current layout, bounds, padding, and scroll size.
- Clamp auto-grow vertical scroll offsets before current-frame painting.
- Add `examples/input_auto_grow_flicker` for manual reproduction.
- Add focused unit coverage for scrollbar geometry and auto-grow offset clamping.

### Before

https://github.com/user-attachments/assets/a7c4c567-8e91-4128-958c-c8490b8ce0fe

### After

https://github.com/user-attachments/assets/3b83ed65-fc34-4b75-8735-dede0b959e2e

## Validation

- `cargo test -p gpui-component test_auto_grow_scroll_offset_is_clamped_to_current_viewport`
- `cargo test -p gpui-component test_editor_scrollbar_layout_uses_current_scroll_size`
- `cargo check -p gpui-component`
- `cargo check -p input_auto_grow_flicker`
- `git diff --check`


crates/ui/src/input/element.rs
```diff
@@ -26,6 +30,55 @@ const FOLD_ICON_WIDTH: Pixels = px(14.);
 const FOLD_ICON_HITBOX_WIDTH: Pixels = px(18.);
 const MAX_HIGHLIGHT_LINE_LENGTH: usize = 10_000;
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+struct EditorScrollbarLayout {
+    bounds: Bounds<Pixels>,
+    scroll_size: Size<Pixels>,
+}
+
+fn editor_scrollbar_layout(
```
huacnlee: Move this method to as EditorScrollbarLayout's method.

crates/ui/src/input/element.rs
```diff
@@ -1605,12 +1675,43 @@ impl Element for TextElement {
...
+            .scroll_size(scrollbar_layout.scroll_size)
+            .into_any_element();
+
+            scrollbar.prepaint_as_root(
+                scrollbar_layout.bounds.origin,
+                scrollbar_layout.bounds.size.into(),
+                window,
+                cx,
+            );
+            Some(scrollbar)
```
huacnlee: Should we can move this part to EditorScrollbarLayout, to avoid add more logic in here.

huacnlee: Like the before `layout_fold_icons`, or other `layout_xxx`

crates/ui/src/input/input.rs
```diff
@@ -283,36 +254,40 @@ impl RenderOnce for Input {
...
 
         div()
             .id(("input", self.state.entity_id()))
             .flex()
             .key_context(crate::input::CONTEXT)
-            .track_focus(&state.focus_handle.clone())
+            .track_focus(&focus_handle)
             .tab_index(self.tab_index)
-            .when(!state.disabled, |this| {
+            .when(!disabled, |this| {
```
huacnlee: Next time when you change code, please avoid to touch code that change not required, to speed up me to review changes.

huacnlee: These changes might inadvertently omit a piece of logic, leading to unnecessary bugs. Is it possible to revert them?

suxiaoshao:
Code review feedback has been addressed. Please re-check when you have a chance.

huacnlee:
1. Scrollbar are not appear now in the auto grow input, we need keep fix it.
2. You not need add another example, there already have one: `cargo run textarea`.

suxiaoshao:
Both points have been addressed: the auto-grow input scrollbar has been restored, and the dedicated input_auto_grow_flicker example has been removed in favor of using cargo run textarea for verification.

huacnlee:
Thank you.

PR #2321, opened 2026-04-30
menu: fix app menu action routing after hover switching

Third-Thing:
**Problem:**
Menu items could sometimes do nothing after switching between top-level menus by hovering. The menu was using the current focus when each popup opened. After the first popup opened, focus moved into that popup. When the user hovered to another menu, the new popup could accidentally remember the old popup as the place to send actions. That old popup was no longer active, so the selected menu item had nowhere useful to go.

**Reproduce:**

1. Select Editor in Story
2. Select any text
3. Click the Help menu
4. Move to the Edit menu and select Cut

Nothing happens.

**Solution:**
The menu bar now remembers the focused area once, when the menu interaction starts. It keeps using that same target while the user switches between menus, and clears it when the menu closes. Popup menus, including submenus and reused popups, now receive this stored target instead of reading focus again.

This keeps menu actions routed to the view that was active before the menu opened, even after hover-switching between menus.

> Implemented-by: Codex GPT-5.5 High
> Reviewed-by: Claude Opus 4.7 High & Human
> Tested-by: Human

**Tests run using Story:**

- Click menu A, click an item in menu A.
- Click menu A, hover menu B, click an item in B.
- Click menu A, hover B, hover C, click an item in C.
- Open with click, switch menus with left/right arrow keys, press Enter on an item.
- Open a menu, press Escape. Confirm focus returns to the original widget.
- Open a menu, click outside it. Confirm focus returns and the menu closes.
- Open a menu from widget 1, dismiss it, focus widget 2, reopen a menu. Confirm actions route to widget 2.
- Use a submenu item, especially after hover-switching top-level menus first.
- Trigger both focus-scoped actions, like copy/paste/delete, and global actions, like about/quit/theme, if present.
- Try from different focused widgets: editor/input, sidebar/tree/list, and no obvious focused widget.

## Checklist

---

Hello - and thank you for this toolkit.

### DIFF
--- a/crates/ui/src/menu/app_menu_bar.rs
+++ b/crates/ui/src/menu/app_menu_bar.rs
@@ -7,7 +7,7 @@ use crate::{
     menu::PopupMenu,
 };
 use gpui::{
-    App, AppContext as _, ClickEvent, Context, DismissEvent, Entity, Focusable,
+    App, AppContext as _, ClickEvent, Context, DismissEvent, Entity, FocusHandle, Focusable,
     InteractiveElement as _, IntoElement, KeyBinding, MouseButton, OwnedMenu, ParentElement,
     Render, SharedString, StatefulInteractiveElement, Styled, Subscription, Window, anchored,
     deferred, div, prelude::FluentBuilder, px,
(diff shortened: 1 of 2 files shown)


madcodelife:
Thank you.

PR #2328, opened 2026-05-01
editor: Add DocumentRangeSemanticTokensProvider (LSP semantic tokens) for the code editor

glani:
## Summary

Adds a `DocumentRangeSemanticTokensProvider` to the code editor — the editor counterpart of the LSP `textDocument/semanticTokens/range` request (and Monaco Editor's provider of the same name). It lets consumers contribute extra syntax highlighting (language servers, syntect, custom DSLs) on top of the built-in tree-sitter highlighter, without replacing it.

It plugs into the existing `editor.lsp.*` provider system, exactly like `document_color_provider`:

```rust
editor.lsp.semantic_tokens_provider = Some(Rc::new(MyProvider));
```

The editor fetches it (debounced) on document change, caches the result as `(range, token-name)` pairs, and resolves names → colors against the active `HighlightTheme` at paint time, so theme switches recolor without a refetch. Tokens compose between tree-sitter (base) and diagnostics (top). Reuses `lsp-types` `SemanticTokens` / `SemanticTokensLegend`.

image

See `examples/markdown.rs` (`MarkerHighlighter`, highlighting `TODO`/`FIXME`/…) for a runnable example.


huacnlee:
The [Syntect](https://github.com/trishume/syntect/) is rather poor; 

I used it in the early stages of implementing CodeEditor. It's not suitable for complex applications.

TreeSitter is currently the best choice, and it's easily extensible. I've already designed ways to extend it; you can see the example of adding tree-sitter-navi in ​​the documentation.

https://github.com/longbridge/gpui-component/blob/01a116a15e9660963a2aa07d0192e38785d8b9ad/crates/story/examples/editor.rs#L56-L66

And TreeSitter is easy to extend a new language.

Therefore, I don't really want to introduce this mechanism, as it would bring more problems. 

First, `styles` method are designed for high-performance scenarios involving highlighting massive amounts of code (200,000 lines or more). To enabling this customize, especially for users with insufficient understanding of editor development, will likely result in poor performance. This requires a high level of knowledge in the code editor domain from the user.

huacnlee:
I think I understand your requirements. Sorry, you wrote too much, but my English isn't very good.

Please look at the implementation of [MonacoEditor](https://microsoft.github.io/monaco-editor/docs.html) to see how its API is designed to support this requirement. We should refer to it.

And then please add an example of this new feature to `examples/editor.rs` so I can test this case when I make changes later.

When you finished that, I will review and accept this feature.

glani:
Hey. 

Revised per your review.

Trait now returns `(Range, SharedString)` — implementors emit scope
names (`"keyword"`, `"comment"`, …); render path resolves them
through the existing `HighlightTheme::style()`. Same vocabulary as
the tree-sitter path. No `&App` on the hot path, so heavy parsing
runs off-thread.

Added example at `crates/story/examples/editor.rs` — highlights
TODO/FIXME/XXX/HACK/NOTE.

Please have a look.

huacnlee:
Thanks for the iterations! I pushed a revision that reshapes this around our existing LSP provider system, so it lands consistently with the rest of the editor. Two changes:

**1. `CustomHighlighter` → `DocumentRangeSemanticTokensProvider`**

Following the Monaco reference, this is now modeled on Monaco's `DocumentRangeSemanticTokensProvider` / the LSP `textDocument/semanticTokens/range` request, and reuses `lsp-types` `SemanticTokens` / `SemanticTokensLegend` (which the editor already depends on).

It plugs into `editor.lsp.*` exactly like `document_color_provider`, instead of a separate `set_*` method:

```rust
editor.lsp.semantic_tokens_provider = Some(Rc::new(MyProvider));
```

The wiring mirrors `document_color_provider` end-to-end: debounced fetch on document change, cached as `(range, token-name)` pairs, then resolved name → color against the active `HighlightTheme` at paint time (so theme switches recolor without a refetch) and composed between tree-sitter (base) and diagnostics (top). The old `InputMode::CodeEditor.custom_highlighter` field and `InputState::set_custom_highlighter` are removed.

**2. Example moved to the markdown example**

`MarkerHighlighter` (TODO/FIXME/XXX/HACK/NOTE) now lives in `examples/markdown.rs` (a `code_editor(Language::Markdown)`), and the markdown fixture gained a "Task markers" section as a test case.

Also merged latest `main` and resolved the conflict. CI is green on all three platforms. Please take a look — if this direction looks good to you I'll get it merged.

glani:
Direction looks good — modeling it on the LSP semanticTokens/range provider and folding it into editor.lsp.* is cleaner.

Thanks for the iteration.

PR #2333, opened 2026-05-04
highlighter: add inline markdown syntax highlighting

Third-Thing:
This PR adds highlighting for common inline Markdown content:

- Bold text
- Italic text
- Strikethrough text
- Inline code
- Link text and link URLs
- Inline LaTeX math spans

Existing styling for links and inline code was left unchanged. Inline LaTeX math currently uses the same visual style as inline code.

| Before | After |
| ------ | ----- |
| Before screenshot | After screenshot |

It also updates the syntax style system so underline and strikethrough styles from themes are shown correctly.

To avoid extra work for plain text, Markdown inline parsing is skipped when the text does not contain inline Markdown markers. This is especially useful for long plain text and CJK text.

This PR only changes inline Markdown highlighting. Code blocks and block quotes are not changed.

## Tests

Added tests for:

- Bold, italic, and nested formatting
- Strikethrough styling
- Inline code
- Link text
- Inline LaTeX math spans
- Plain text skipping inline parsing
- Paragraph boundaries for inline parsing

Also tested the Markdown story on Linux.

## Breaking Changes

- Added a new `Strikethrough` variant to the public `FontStyle` enum used by theme syntax styles. Code that matches this enum exhaustively must add a new arm.
- Added a new optional `strikethrough`  field on the public `SyntaxColors` struct. Rust code that constructs this struct with all fields listed must add the new field or use `Default`.

## Notes

While testing mixed CJK text in the Markdown story, a rendering delay was found when the default Linux monospace font required CJK font fallback. Using a font with CJK coverage, such as Sarasa Mono SC, removed the delay. This appears to be related to upstream font fallback behavior. A separate optimization for longest line calculation will be handled in another PR.

---

> Implemented by: Codex GPT-5.5
> Ideas/suggestions/reviews by: Claude Opus 4.7 and Human

## Checklist

- [x] I have read the [CONTRIBUTING](../CONTRIBUTING.md) document and followed the guidelines.
- [x] Reviewed the changes in this PR and confirmed AI generated code (If any) is accurate.
- [x] Passed `cargo run` for story tests related to the changes.
- [x] Tested Linux performance

PR #2338, opened 2026-05-05
Revert "highlighter: add inline markdown syntax highlighting"

huacnlee:
Reverts #2333, cause of there have some special hard code for Markdown.

Sorry @Third-Thing, after I revert #2333 again, that implementation was not good. 

We should avoid to add hard code for `markdown` in highlighter.

You can try again to find a new way to solve that issue, we are happy to help to to review and push merge as as soon as possible.


Third-Thing:
Do you mean the optimization code?

```rust
fn should_parse_injection_layer(...) -> bool {
    if language_name.as_ref() != "markdown_inline" {
        return true;
    }
    ranges.iter().any(|range| markdown_inline_range_has_trigger(text, ...))
}

fn markdown_inline_range_has_trigger(text: &Rope, range: Range<usize>) -> bool {
    // hardcoded: `*`, `_`, `` ` ``, `[`, `]`, `(`, `)`, `<`, `>`, `!`, `~`, `$`
}
```

huacnlee:
Yes

PR #2340, opened 2026-05-05
input: Fix folding perf regression

zanmato:
## Description

#2260 fixed a bug (thanks!) but regressed in performance for large files.
This moves the folding to the background since the tree is available there. Note that this makes highlighting big files a bit slower since we wait for folding, but it doesn't disrupt the main thread at least. Perhaps there are performance gains in the collect foldable nodes functions?

### Before (~15s on main thread)
Screenshot From 2026-05-05 13-03-13

### After (on worker thread)
Screenshot From 2026-05-05 13-05-25

## How to Test

1. Generate test file `(echo '['; for i in $(seq 1 9999); do echo "{\"id\": $i, \"name\": \"user_$i\"},"; done; echo '{"id": 10000, "name": "user_10000"}'; echo ']') > test.json`
2. Disable/comment `lint_document` and `document_color_provider` in `editor.rs`
3. Run editor `cargo run --release --example editor`
4. Test folding

## Checklist

### DIFF
--- a/crates/ui/src/input/mode.rs
+++ b/crates/ui/src/input/mode.rs
@@ -16,6 +16,7 @@ pub(super) struct PendingBackgroundParse {
     pub parse_task: Rc<RefCell<Option<Task<()>>>>,
     pub language: SharedString,
     pub text: Rope,
+    pub is_folding: bool,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


huacnlee:
Thanks

PR #2342, opened 2026-05-05
input: Fix the InputEdit sent to tree-sitter when text is replaced

Third-Thing:
Tree-sitter needs the old end position to come from the text before the edit. The previous code measured it from the text after the edit. For deletions at the end of the text, this could report the edit as if no old text was removed.

Example:

- Old text: `"[1,2]"`
- New text: `"[1,2"`
- Deleted range: `4..5`
- Correct `old_end_byte`: `5`
- Old code could send `old_end_byte`: `4`

That tells tree-sitter the edit is empty, so tree-sitter can keep old ranges past the end of the new text.

## Why

Tree-sitter documents that incremental parsing requires editing the old tree so it matches the source change:

- `ts_parser_parse`: the old tree must be edited in a way that exactly matches the source code changes: https://docs.rs/tree-sitter/latest/tree_sitter/ffi/fn.ts_parser_parse.html
- `ts_tree_edit`: the edit must describe byte offsets and row/column positions: https://docs.rs/tree-sitter/latest/tree_sitter/ffi/fn.ts_tree_edit.html
- Tree-sitter editing docs show `TSInputEdit` has both `old_end_byte` and `new_end_byte`: https://tree-sitter.github.io/tree-sitter/using-parsers/3-advanced-parsing.html#editing

## What Changed

- Pass the pre-edit text into `update_highlighter`.
- Build `InputEdit` with:
  - old offsets and old positions from the pre-edit text
  - new end offset and new end position from the post-edit text

Added tests for end-of-text deletion.

I checked the new tree-sitter included-range test against the old edit calculation. It failed as expected:

```text
left: 5
right: 4
```

This shows tree-sitter kept an old range ending at byte `5` after the new text ended at byte `4`.

---

> Implemented-by: Codex GPT-5.5
> Reviewed-by: Claude Opus 4.7 and Human

### DIFF
--- a/crates/ui/src/input/mode.rs
+++ b/crates/ui/src/input/mode.rs
@@ -224,6 +224,7 @@ impl InputMode {
     pub(super) fn update_highlighter(
         &mut self,
         selected_range: &Range<usize>,
+        old_text: &Rope,
         text: &Rope,
         new_text: &str,
         force: bool,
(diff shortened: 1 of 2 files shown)


huacnlee:
image

Need more test.

huacnlee:
Oh, that was exist. You can ignore it.

huacnlee:
Thank you.

huacnlee (approved):

PR #2344, opened 2026-05-06
highlighter: Support fixed-language injection layers

Third-Thing:
This is the foundational work from the previous [inline markdown syntax highlighting PR](https://github.com/longbridge/gpui-component/pull/2339). It is now framed in terms of what it solves independently from inline markdown highlighting.

---

**Problem:**

HTML files can contain CSS and JavaScript inside style and script blocks. Before this change, those inner languages were not parsed and highlighted as their own languages. This meant CSS and JavaScript code inside HTML appeared with incomplete or plain highlighting.

This PR makes the highlighter recognize those fixed inner language regions, so CSS inside style blocks and JavaScript inside script blocks receive the correct syntax highlighting.

**Solution:**

- The highlighter now reads all injection rules for a language, not only the special combined case.

- It builds separate parsed layers for fixed inner languages, such as CSS inside style blocks and JavaScript inside script blocks.

- It keeps the existing behavior for cases that need multiple regions parsed together, such as HTML regions inside PHP.

- It reuses parsed inner layers when their language and text ranges have not changed.

- It adds tests that show CSS and JavaScript are highlighted correctly inside HTML.

**Tests:**

- The added failing tests now pass.
- I have manually tested `gpui-component-story --example editor` for intended behavior, edge cases and side effects.

---

> Implemented-by: Codex GPT-5.5
> Reviewed-by: Claude Opus 4.7 and Human

### DIFF
--- a/crates/ui/src/highlighter/highlighter.rs
+++ b/crates/ui/src/highlighter/highlighter.rs
@@ -483,6 +483,21 @@ impl SyntaxHighlighter {
         tree: &Tree,
         text: &Rope,
     ) -> Vec<InjectionLayer> {
+        fn sort_ranges(ranges: &mut [tree_sitter::Range]) {
+            ranges.sort_unstable_by(|a, b| {
+                a.start_byte
+                    .cmp(&b.start_byte)
+                    .then_with(|| a.end_byte.cmp(&b.end_byte))
+            });
+        }
+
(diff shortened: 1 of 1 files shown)


Third-Thing:
@huacnlee Thanks for working with me on this.

PR #2345, opened 2026-05-06
Notify when a notification dismisses

dunkmann00:
When a notification dismissed itself after the default delay in my app, it was not totally disappearing. It would turn semi-transparent, but not totally disappear until something else happened that triggered the view to redraw. I added the `notify` call in the dismiss handler and it worked fine again.

## How to Test

Push a notification, have nothing else happen so the root view doesn't get redrawn by some other trigger.

## Checklist


madcodelife:
Thanks.

PR #2350, opened 2026-05-08
text: Enable mixed bold/italic markdown rendering

### LINKED ISSUE #2120: Markdown: nested bold+italic (***text***) loses inner formatting
## Bug Description

When rendering markdown with combined bold and italic formatting (`***bold and italic***`), only the outermost mark is applied — the inner formatting is lost. Bold (`**text**`) and italic (`*text*`) work correctly on their own, but not when nested.

## Root Cause

In `crates/ui/src/text/format/markdown.rs`, the `parse_paragraph` function handles `Node::Strong` and `Node::Emphasis` by:

1. Creating a `child_paragraph`
2. Recursively calling `parse_paragraph` on children (which pushes marked `InlineNode`s into `child_paragraph`)
3. **Discarding `child_paragraph`** and creating a new `InlineNode` from only the returned text string, applying just the current mark

For `***bold and italic***`, the markdown AST produces:

```
Strong
  └── Emphasis
        └── Text("bold and italic")
```

When `Strong` processes its children, `Emphasis` pushes an italic-marked node into `child_paragraph` — but `Strong` ignores that and creates a new node with only `bold()`. The italic mark is lost.

**Relevant code** (lines ~76-92):

```rust
Node::Emphasis(val) => {
    let mut child_paragraph = Paragraph::default();
    for child in val.children.iter() {
        text.push_str(&parse_paragraph(&mut child_paragraph, &child, cx));
    }
    paragraph.push(
        InlineNode::new(&text).marks(vec![(0..text.len(), TextMark::default().italic())]),
    );
}
Node::Strong(val) => {
    let mut child_paragraph = Paragraph::default();
    for child in val.children.iter() {
        text.push_str(&parse_paragraph(&mut child_paragraph, &child, cx));
    }
    paragraph.push(
        InlineNode::new(&text).marks(vec![(0..text.len(), TextMark::default().bold())]),
    );
}
```

## Suggested Fix

When the children produce `InlineNode`s with marks (in `child_paragraph`), merge those marks with the current formatting mark rather than discarding them. For example, when `Strong` finds that its child `Emphasis` already produced italic-marked nodes, the result should carry both `bold` and `italic`.

## Reproduction

```markdown
**bold** works
*italic* works
***bold and italic*** — only shows bold (or italic), not both
```

## Environment

- `gpui-component` version: 0.5.1

## Impact

Low — workaround is to avoid `***text***` syntax. But it's standard markdown that users expect to work.

Third-Thing:
The markdown preview did not show text correctly when bold and italic were used together. For example, text marked as both bold and italic could lose one of those styles.

The solution keeps the existing text styles when nested markdown is parsed, then adds the outer style to the same text. This lets combined styles, such as bold plus italic, render correctly in the preview.  (Closes #2120)

**Tested in `gpui-component-story --example markdown`**
```
_italic_ *italic*
**bold** __bold__
***bold and italic*** ___bold and italic___
_**bold and italic**_ *__bold and italic__*
*italic with **bold*** _italic with __bold___
_italic **with** bold_ *italic __with__ bold*
**bold with *italic*** __bold with _italic___
__bold *with* italic__ **bold _with_ italic**
```
Screenshot_20260508_122144

---

Implemented-by: Codex GPT-5.5
Reviewed-by: Claude Opus 4.7 and Human

### DIFF
--- a/crates/ui/src/text/format/markdown.rs
+++ b/crates/ui/src/text/format/markdown.rs
@@ -1,3 +1,5 @@
+use std::ops::Range;
+
 use gpui::SharedString;
 use markdown::{
     ParseOptions,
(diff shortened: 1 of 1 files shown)


madcodelife:
Thanks! I tested the styles inside the markdown reference are still not preserved. Could you handle this in the same PR?

image

Third-Thing:
```
- [**one _two_ three**][r]
- [*one __two__ three*][r]
- [***one two three***][r]

[r]: https://example.com
```
Screenshot_20260509_211941

madcodelife:
Thanks.

PR #2354, opened 2026-05-11
docs: Update data-table docs to return Column by owned

HuaGu-Dragon:
https://github.com/longbridge/gpui-component/blob/bd6c4dfafff1b3ce2f5634e487567697aac8b8bb/crates/ui/src/table/delegate.rs#L26


madcodelife:
Thanks.

PR #2365, opened 2026-05-13
docs: Fix column return and reformat

HuaGu-Dragon:
See #2354. My apologies — I forgot to remove the reference. The Chinese version is fine.


madcodelife:
Thanks.

PR #2371, opened 2026-05-13
text_view: Improve markdown preview responsiveness during rapid text updates

Third-Thing:
## Problem

The markdown preview was applying every completed parse result, even when newer edits had already arrived. When typing or deleting quickly, slow parses could pile up and the preview would render old versions before catching up.

## Solution

While parsing is still running, the preview keeps showing the last accepted parsed document. When parsing finishes, the result is accepted only if its update number is still current. If it is current, the preview is notified and renders the new document. If it is stale, the result is discarded and does not trigger a render.

After a parse finishes, the worker now drains queued edits and merges them into one latest update before parsing again. Before this change, queued edits could each trigger their own parse. During rapid input, this reduces both parsing and rendering work.

### Improvements

- Fewer follow-up parses when edits pile up
- Fewer preview renders because stale parse results are ignored

---

Implemented-by: Codex GPT-5.5
Reviewed-by: Claude Opus 4.7 and Human


Third-Thing:
Before
[before-render-improvement.webm](https://github.com/user-attachments/assets/c5967e62-f7d6-4fcb-a8dd-53ba5aa030b9)

After
[after-render-improvement.webm](https://github.com/user-attachments/assets/d3b35e0d-5374-4b66-858c-8764d50c1d3b)

madcodelife:
After #2400 was merged, I tested the scenario shown in the video(on macOS), and it looks very smooth now.

Do you mind giving it a try?

Third-Thing:
Yes that helps, which is not surprising. I flagged code blocks as a performance problem, but this more general optimization was found during investigation.

I just tested a markdown file with 637,707 characters and the rendering performance is still very slow. So for large files this optimization would still be useful.
[Screencast_20260531_231632.webm](https://github.com/user-attachments/assets/d9e49067-b0c8-4448-985c-f28d2e5f2a79)

PR #2382, opened 2026-05-18
tooltip: Fix to hide Tooltip immediately on mouse down.

### LINKED ISSUE #2381: Tooltip remains visible after its trigger is removed
## Description

A tooltip can remain visible after the element that triggered it is removed from the UI tree.

This happens when a tooltip-bearing control performs an action that removes itself, such as a delete/remove button. The trigger disappears, but the tooltip overlay remains rendered and there is no longer a trigger element to emit `hover(false)`, so the tooltip cannot be dismissed through the normal hover lifecycle.

## Environment

- GPUI: zed git revision `832c17e8` from the current lockfile
- GPUI Component: `0.5.2` / current `main`
- Platform: macOS

## Steps to Reproduce

1. Run the Story Gallery.
2. Open the Tooltip story.
3. In `Tooltip trigger removed on click`, hover the `Remove me` button until its tooltip appears.
4. Click `Remove me`.

## Screenshots

https://github.com/user-attachments/assets/9a88bde2-402c-4b3a-a269-61e4491b5533

## Expected

The tooltip should be hidden when its trigger element is removed, or at least be dismissed immediately after the click removes the trigger.

## Actual

The `Remove me` button disappears, but the tooltip remains visible. Since the trigger is gone, it does not emit `hover(false)`, and the Root-level tooltip overlay keeps rendering the stale tooltip content.

## Notes

A reproduction case has been added locally to the Tooltip story: `Tooltip trigger removed on click`.

suxiaoshao:
Closes #2381

## Description

Fix stale managed tooltips when a tooltip trigger removes itself on click. Managed tooltips are rendered by the Root overlay, so if the trigger is removed before the normal hover-leave path runs, the old tooltip can remain visible and cannot be dismissed by moving away from the removed trigger.

This change clears the active tooltip immediately on trigger mouse down, before the click handler mutates the UI tree. The normal hover-enter delay and hover-leave grace period are otherwise preserved.

This follows the shadcn/base-ui tooltip behavior. shadcn/ui delegates tooltip behavior to Base UI in its base implementation, and Base UI defaults tooltip triggers to `closeOnClick = true`. Base UI cancels delayed hover opens on trigger `onPointerDown` / `onClick`, and uses `useDismiss` with `referencePress` to dismiss an already-open tooltip when the trigger is pressed:

```tsx
const {
  closeOnClick = true,
} = componentProps;

onPointerDown(event) {
  store.set("closeOnClick", closeOnClick);
  if (closeOnClick && !store.select("open")) {
    store.cancelPendingOpen(event.nativeEvent);
  }
}

onClick(event) {
  if (closeOnClick && !store.select("open")) {
    store.cancelPendingOpen(event.nativeEvent);
  }
}

useDismiss(floatingRootContext, {
  referencePress: () => store.select("closeOnClick"),
});
```

## Screenshot

Before

https://github.com/user-attachments/assets/9a88bde2-402c-4b3a-a269-61e4491b5533

After

https://github.com/user-attachments/assets/7f63c5f3-36a5-407d-8dcb-c42e3402b97c

## How to Test

- `cargo test -p gpui-component tooltip`
- `cargo check -p gpui-component-story`
- `rustfmt --check crates/ui/src/tooltip.rs crates/story/src/stories/tooltip_story.rs`
- `git diff --check`

Manual verification steps:

1. Run `cargo run`.
2. Open the Tooltip story.
3. Hover `Remove me` until the tooltip appears.
4. Click `Remove me`.
5. Expected: the button disappears and the tooltip disappears with it.

## Checklist

### DIFF
--- a/crates/story/src/stories/tooltip_story.rs
+++ b/crates/story/src/stories/tooltip_story.rs
@@ -1,6 +1,7 @@
 use gpui::{
     App, AppContext, Context, Entity, Focusable, InteractiveElement, KeyBinding, ParentElement,
     Render, StatefulInteractiveElement as _, Styled, Window, actions, div,
+    prelude::FluentBuilder as _,
 };
 
 use gpui_component::{
@@ -26,6 +27,7 @@ pub fn init(cx: &mut App) {
 
 pub struct TooltipStory {
     focus_handle: gpui::FocusHandle,
+    removable_button_visible: bool,
 }
 
 impl TooltipStory {
@@ -36,6 +38,7 @@ impl TooltipStory {
     fn new(_: &mut Window, cx: &mut Context<Self>) -> Self {
         Self {
             focus_handle: cx.focus_handle(),
+            removable_button_visible: true,
         }
     }
 }
@@ -68,7 +71,7 @@ impl Render for TooltipStory {
     fn render(
         &mut self,
         _: &mut gpui::Window,
-        _cx: &mut gpui::Context<Self>,
+        cx: &mut gpui::Context<Self>,
     ) -> impl gpui::IntoElement {
         v_flex()
             .w_full()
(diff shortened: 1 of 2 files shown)


huacnlee:
Thanks.

PR #2396, opened 2026-05-22
fix(linux): correct CSD window border resize cursor and hit-testing

gaoyia:
## Summary

Fixes stale resize cursors on Linux client-side decorations (CSD): the pointer can keep showing a horizontal resize shape after leaving a left/right edge, even when hovering over a top/bottom edge where drag behavior is already correct.

## Symptoms

On Linux with `WindowDecorations::Client` (via `Root` + `window_border`):

1. Hover over a left or right border → cursor shows **horizontal** resize.
2. Move away from that edge, then hover over the **bottom** border.
3. **Cursor still shows horizontal resize**, but clicking and dragging performs a **vertical** resize as expected.

Cursor feedback and actual resize direction can disagree, especially when moving between edges without leaving the window.

## Root cause

Resize cursor updates live in `window_border.rs`, inside the `canvas` **paint** callback. Two issues combine:

1. **No reset when leaving an edge**   When `resize_edge()` returned `None`, the paint callback returned early and never called `set_cursor_style(CursorStyle::default(), ...)`. The last resize cursor (e.g. horizontal) remained on screen.

2. **Cursor update path ≠ resize click path**  
   - **Cursor**: updated only in the canvas paint callback (full-window hitbox; GPUI requires `set_cursor_style` during paint).
   - **Resize on click**: handled on the outer padded `#window-backdrop` via `on_mouse_down` + `start_window_resize`, using the **current** mouse position.

   Click/drag can therefore be correct while the displayed cursor is stale from a previous edge.

   The padding strip (`shadow_size`) also sits structurally below the inner content layer; without a refresh trigger, paint may not re-run often enough to update the cursor in the border hit zone.

## Fix

- Extract `update_resize_cursor()` and `cursor_style_for_resize_edge()`.
- In canvas paint: when not on a resize edge, set `CursorStyle::default()` instead of returning early.
- On the padded backdrop: `on_mouse_move` → `window.refresh()` so paint-phase cursor updates run reliably in the border padding area.

## Test plan

- [ ] Linux + Wayland/X11, app using `Root` with CSD (`WindowDecorations::Client`).
- [ ] Hover left/right edge → horizontal cursor; move to interior → default arrow.
- [ ] From left edge, move to bottom edge → vertical cursor before click.
- [ ] Click-drag on each edge/corner: cursor shape matches resize direction.
- [ ] No panic on mouse move (`set_cursor_style` must stay in paint; do not call it from `on_mouse_event` handlers).

## Changes

```diff
+ Introduce `RESIZE_HIT_SIZE` (default 4px) and `WindowBorder::resize_hit_size()`.
```

## Checklist

### DIFF
--- a/crates/ui/src/window_border.rs
+++ b/crates/ui/src/window_border.rs
@@ -187,6 +187,7 @@ impl RenderOnce for WindowBorder {
                                     blur_radius: visual_shadow / 2.,
                                     spread_radius: px(0.),
                                     offset: point(px(0.0), px(0.0)),
+                                    inset: false,
                                 }])
                             }),
                     })
--- a/crates/ui/src/window_border.rs
+++ b/crates/ui/src/window_border.rs
@@ -1,10 +1,9 @@
 // From:
 // https://github.com/zed-industries/zed/blob/56daba28d40301ee4c05546fadb691d070b7b2b6/crates/gpui/examples/window_shadow.rs
 use gpui::{
-    AnyElement, App, Bounds, CursorStyle, Decorations, Edges, Hitbox, HitboxBehavior, Hsla,
-    InteractiveElement as _, IntoElement, MouseButton, ParentElement, Pixels, Point, RenderOnce,
-    ResizeEdge, Size, Styled as _, Tiling, Window, canvas, div, point, prelude::FluentBuilder as _,
-    px,
+    AnyElement, App, CursorStyle, Decorations, Edges, Hsla, InteractiveElement as _,
+    IntoElement, MouseButton, ParentElement, Pixels, Point, RenderOnce, ResizeEdge, Size,
+    Styled as _, Tiling, Window, div, point, prelude::FluentBuilder as _, px,
 };
 
 use crate::ActiveTheme;
@@ -109,6 +108,7 @@ impl RenderOnce for WindowBorder {
         };
         let resize_hit_size = self.resize_hit_size;
         window.set_client_inset(shadow_size);
+        let window_size = window.window_bounds().get_bounds().size;
 
         div()
             .id("window-backdrop")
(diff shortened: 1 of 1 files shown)


crates/ui/src/window_border.rs
```diff
@@ -155,6 +146,10 @@ impl RenderOnce for WindowBorder {
                     .when(!tiling.bottom, |div| div.pb(shadow_size))
                     .when(!tiling.left, |div| div.pl(shadow_size))
                     .when(!tiling.right, |div| div.pr(shadow_size))
+                    .on_mouse_move(move |_, window, _| {
```
madcodelife: This handler calls `window.refresh()` on every mouse-move event, which triggers a full window repaint each time. PR #617 previously removed this exact pattern from this file for performance reasons.

Is it possible to add some checks so that it is only triggered when a refresh is needed?

gaoyia: Done. Instead of gating window.refresh() on mouse move, I replaced the canvas + refresh path with per-edge hit layers that use .cursor(). GPUI updates the cursor when the hovered hitbox changes, so we don’t need a full repaint on every mouse_move—same goal as #617, without the one-frame lag when switching between adjacent resize edges.

madcodelife:
Nice rework, the cursor handling looks much cleaner now.

One quick question: with `resize_hit_zones` now owning all resize hit-testing and `start_window_resize` calls, is the outer `on_mouse_down` still needed?

https://github.com/longbridge/gpui-component/blob/13115c2e84f269d870c7434a29f1e6a0edcd1a53/crates/ui/src/window_border.rs#L139-L154

https://github.com/longbridge/gpui-component/blob/13115c2e84f269d870c7434a29f1e6a0edcd1a53/crates/ui/src/window_border.rs#L243-L246

gaoyia:
this PR actually flips it the other way: zones no longer
call start_window_resize at all. They only own cursor styling now.
All resize initiation goes through the backdrop handler via resize_edge(),
so the outer on_mouse_down is not just needed, it's the only path.

fix: when the window was tiled on all sides, set_client_inset was being cleared along with the visual padding. That caused compute_outer_size to double-count the shadow on the first resize after restoring, making the window jump larger. Now platform_inset stays at the full shadow size so the platform inset is stable regardless of tiling state.

gaoyia:
I kept the backdrop on_mouse_down and removed the per-zone handlers based on the AI reviewer's suggestion. The thinking is: while zones look like the obvious place to handle resize clicks, they can easily be covered by content that stretches to the window edge (like a full-width toolbar or a scrollbar). When that happens the zones never see the click and resizing silently breaks. The backdrop sits underneath everything, so clicks that aren't consumed bubble up to it — it acts as a reliable fallback. It's a little counterintuitive at first, but it makes resize handling way more robust across different window layouts.

madcodelife:
Thank you!

PR #2397, opened 2026-05-23
tabs: Fix segmented indicator flicker and add a new section to tab story

### LINKED ISSUE #2395: Tab suffix renders outside the inner background in Segmented variant
## Description

When using TabBar with the Segmented variant, the suffix element (for example, a close button) may render outside of the inner content background (inner_bg).

The issue does not occur on the initial render, but can appear after creating new tabs dynamically.

## Environment

- GPUI: latest commit
- GPUI Component: latest commit
- Platform: Windows 11

### Incorrect Rendering
Image

### Expected Rendering
Image

BeratHundurel:
Closes #2395

## Description
Fixes segmented tab suffix flicker when tabs are added or switched by rendering a fallback inner background until the indicator is ready and deferring per‑tab inner styling to the indicator. Adds a simple story repro with dynamic segmented tabs and suffix buttons so the issue can be tested reliably.

### Before
Here we can see Tab 3’s inner background grow on the next paint after it’s added. A similar effect happens when switching tabs with clicks.

https://github.com/user-attachments/assets/6d9822bf-76af-44b4-944a-94e971498956

### After
And now it's consistent and flicker is not happening. 

https://github.com/user-attachments/assets/82e60ab4-162f-4199-a718-ced3eef66a11

## How to Test
1. Run `cargo run`.
2. Open the **Tabs** story.
3. Use **Segmented Tabs (Dynamic)**.
4. Click **Add Tab** and **Remove Last** repeatedly.
5. Confirm the selected tab’s inner background stays correct and the suffix stays inside during/after the add animation.

## Checklist

### DIFF
--- a/crates/story/src/stories/tabs_story.rs
+++ b/crates/story/src/stories/tabs_story.rs
@@ -4,7 +4,7 @@ use gpui::{
 };
 
 use gpui_component::{
-    ActiveTheme as _, IconName, Selectable as _, Sizable, Size,
+    ActiveTheme as _, Icon, IconName, Selectable as _, Sizable, Size,
     button::{Button, ButtonGroup, ButtonVariants},
     checkbox::Checkbox,
     h_flex,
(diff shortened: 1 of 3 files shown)


madcodelife:
Thank you, I made a few small style adjustments.

PR #2400, opened 2026-05-25
text: Improve markdown code block rendering performance

BeratHundurel:
## Description

This changes code block syntax highlighting to compute styles lazily and reuse `SyntaxHighlighter` instances through a thread-local language cache. This avoids repeated highlighter construction during markdown parsing/render setup for large files with many code blocks.

It also avoids repeatedly recombining visible input highlight ranges when the ranges are already disjoint.

I also tested #2371 against the same case, but it did not help with the slowdown I was seeing.

I first noticed the issue in my side project, then reproduced the same behavior in the example markdown editor. In local testing, Task Manager showed lower CPU usage while editing with this implementation, memory usage stayed about the same.

Adds regression coverage for code block equality and cached highlighter refresh behavior when a language is registered after an initial fallback.

AI assistance was used to help and prepare this PR. I manually reviewed the resulting code and tested the performance difference before opening the PR.

## Test File

I used this AI-generated markdown file to reproduce the issue and test the change:

[refactor-opportunities.md](https://github.com/user-attachments/files/28234513/refactor-opportunities.md)

### Before

<video src="https://github.com/user-attachments/assets/dbcc45ab-f371-40c2-b6b5-1336fee995fe" controls></video>

### After

<video src="https://github.com/user-attachments/assets/4e3c0754-55b5-4d54-9eb6-a4261f940ad9" controls></video>

## Tests

- `cargo test -p gpui-component code_block_`
- `cargo test -p gpui-component input::`
- `cargo test -p gpui-component text::`

## Checklist

- [x] Reviewed the changes in this PR and confirmed AI-generated code, if any, is accurate.
- [x] Tested macOS, Windows, and Linux platform performance, if the change is platform-specific.

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -1260,7 +1260,8 @@ impl TextElement {
 
         let mut styles = Vec::with_capacity(visible_buffer_lines.len());
 
-        // Helper to flush a contiguous range of lines
+        // Helper to flush a contiguous range of lines. These ranges are disjoint,
+        // so appending avoids repeatedly cloning and recombining prior styles.
         let flush_range = |start_line: usize, end_line: usize, skip: bool, styles: &mut Vec<_>| {
             let byte_start = text.line_start_offset(start_line);
             let byte_end = if is_multi_line {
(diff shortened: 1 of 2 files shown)


madcodelife:
Thank you, it looks great!

PR #2410, opened 2026-05-29
editor: Add scroll_beyond_last_line, cursor_surrounding_lines option

glani:
## Problem

In code-editor mode, `InputState` hardcodes two scroll behaviors behind a
single shared constant (`BOTTOM_MARGIN_ROWS`):

- **Scroll past last line** — how far below the last line of content the
  user can scroll (the empty band at the bottom of the editor).
- **Cursor surrounding lines** — how many lines the cursor is kept clear
  of the viewport's top and bottom edges while moving.

These are two distinct concepts that happen to share one default, and
neither can be tuned without patching the crate. Different editors want
different values: a full-pane editor usually wants a small trailing band
so the bottom of the pane doesn't read as missing content, while a
compact or embedded editor usually wants none.

## Solution

Expose both as separate, optional settings on `InputState`, the same way
VSCode (`editor.scrollBeyondLastLine`, `editor.cursorSurroundingLines`)
and JetBrains IDEs split them:

```rust
pub fn scroll_past_last_line_rows(self, rows: Option<usize>) -> Self;
pub fn set_scroll_past_last_line_rows(&mut self, rows: Option<usize>, ...);

pub fn cursor_surrounding_lines(self, lines: Option<usize>) -> Self;
pub fn set_cursor_surrounding_lines(&mut self, lines: Option<usize>, ...);
```

Both default to `None`, which preserves the current behavior exactly: when
unset, the paint path falls back to the existing `BOTTOM_MARGIN_ROWS`
constant. This is a strict addition — no change to existing behavior.

## Tests

Added coverage for both knobs: `None` reproduces the current layout, and
explicit values change the trailing band and cursor padding as expected.

image

### DIFF
--- a/crates/story/examples/editor.rs
+++ b/crates/story/examples/editor.rs
@@ -76,6 +76,8 @@ pub struct Example {
     soft_wrap: bool,
     show_whitespaces: bool,
     folding: bool,
+    scroll_beyond_last_line: Option<usize>,
+    cursor_surrounding_lines: Option<usize>,
     lsp_store: ExampleLspStore,
     _subscriptions: Vec<Subscription>,
     _lint_task: Task<()>,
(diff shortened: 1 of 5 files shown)


huacnlee:
Thank you.

PR #2411, opened 2026-05-29
input: add editor.gutter.background theme key

glani:
## Problem

The line-number gutter is painted with the same color as the editor text
background (`editor_background()`), and there's no way for a theme to give
the gutter its own background. Some themes (and some editor UIs) want the
gutter visually distinct from the text area — e.g. a slightly different
shade so the line-number column reads as separate chrome.

## Solution

Add an optional `editor.gutter.background` key to the highlight theme:

- `HighlightThemeStyle` gains `editor_gutter_background: Option<Hsla>`,
  parsed via the same `#[serde(rename = "...")]` pattern as the
  surrounding `editor.*` keys.
- The gutter paint in `input/element.rs` reads this field first and falls
  back to `editor_background()` when it's unset.

Because the field is optional and defaults to the existing editor
background, themes that don't define the key render exactly as before —
this is a strict addition with no behavior change.

### DIFF
--- a/crates/ui/src/highlighter/registry.rs
+++ b/crates/ui/src/highlighter/registry.rs
@@ -427,6 +427,10 @@ pub struct HighlightThemeStyle {
     pub editor_active_line_number: Option<Hsla>,
     #[serde(rename = "editor.invisible")]
     pub editor_invisible: Option<Hsla>,
+    /// Optional background color for the gutter (line-number column).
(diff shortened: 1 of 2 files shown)


huacnlee:
Have you done with this PR? Then we can merge it.

glani:
yes. I have a list of small improvements for the editor -)

PR #2414, opened 2026-05-29
assets: Fix dependency of the `ui` crate on `assets` crate

mike-marcacci:
## Description

Previously, generation of default icons escaped the ui crate's directory, assuming the presence of `../assets/assets/icons`. This broke any project using cargo vendor, nix, sandboxed builds, etc.

It might be better to create a separate `icons` crate that holds the SVGs and generates the default icons, but that's a more opinionated change.

This simply updates the macro to support an environment variable based path, and uses cargo links to publish the *real* path to the asset crate's icons.

This is the correct pattern in cargo, but isn't crazy common, hence verbose comments explaining the indirection. [The use of `package.links`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field) is most typically used for [native libraries](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key) but also works perfectly for this case.

## Screenshot

This does not make any visual changes.

## Break Changes

This does not introduce breaking changes.

## How to Test

All existing tests should work without modification. Projects using gpui-component should now build with vendored/sandboxed/nix pipelines.

## AI Assistance

Claude Code wrote all substantial portions of this fix, and I did not see anything that needed revision. Fixed behavior was manually validated in the consuming app.

### DIFF
--- a/crates/macros/src/lib.rs
+++ b/crates/macros/src/lib.rs
@@ -101,13 +101,13 @@ fn pascal_case(filename: &str) -> String {
 ///
 /// ```ignore
 /// // Literal path (relative to the calling crate's CARGO_MANIFEST_DIR)
-/// icon_named!(MyIcons, "icons");
+/// icon_named!(IconName, "icons");
 ///
 /// // Env-var reference (resolved at macro expansion time)
 /// icon_named!(IconName, "$GPUI_COMPONENT_DEFAULT_ICONS_DIR");
 ///
 /// // With custom derives
-/// icon_named!(MyIcons, "icons", [Debug, Copy, PartialEq, Eq]);
+/// icon_named!(IconName, "icons", [Debug, Copy, PartialEq, Eq]);
 /// ```
 #[proc_macro]
 pub fn icon_named(input: TokenStream) -> TokenStream {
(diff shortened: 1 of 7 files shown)


crates/macros/src/lib.rs
```diff
@@ -82,18 +82,32 @@ fn pascal_case(filename: &str) -> String {
...
+///   `gpui-component-assets` without a sibling-crate reference, which would
+///   otherwise break `cargo vendor` and `cargo publish`.
+///
 /// # Example
 ///
 /// ```ignore
-/// // Basic usage (derives IntoElement, Clone by default)
-/// icon_named!(IconName, "../assets/assets/icons");
+/// // Literal path (relative to the calling crate's CARGO_MANIFEST_DIR)
+/// icon_named!(MyIcons, "icons");
```
huacnlee: > You are not enable "Allow maintainer to edit".

Would here we can keep IconName? Why need change this name?

mike-marcacci: No good reason: it was just distinguishing between the "default" icons and "custom" icons. I'll fix that to be more consistent.

PR #2438, opened 2026-06-04
settings: Added a way to add additional search keywords

Libadoxon:
## Description
This adds a field for both SettingsItem::{Item, Element} to allow additional search aliases for that item. This can be used to either provide more search terms for a field, e.g. having a field titled "Enable Two-factor auth" aliased with "MFA". Another use case is to allow custom elements to provide their own search terms as they don't have/want a title/description.

## Screenshot
N/A

## How to Test

`cargo clippy` + verified behavior with example in the settings story 

## Checklist


madcodelife:
I have renamed `search_aliases` to `keywords` and added docs.

Thank you!

PR #2441, opened 2026-06-06
input: Expose API to`set_scroll_offset` method.

hlcfan:
## Description

Expose the API to schedules the editor's viewport scroll offset on the next layout pass. So user can call `defer_scroll_offset` to scroll to an offset when restoring an editor's scroll offset.

## Changes

Add new API `defer_scroll_offset(offset, cx)` to `InputState`.

## How to Test

Replace line https://github.com/longbridge/gpui-component/blob/dadfca97fec7221acf3ce7047bccdc1eac0506b9/crates/story/examples/editor.rs#L802 with 
```rust
let line_height = state.line_height().unwrap_or(window.line_height());
let scroll_offset = state.scroll_offset();
state.defer_scroll_offset(
    point(
        scroll_offset.x,
        -(line_height * position.line as f32),
    ),
    cx,
);
```

## Checklist

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -31,7 +31,6 @@ use super::{
 };
 use crate::Size;
 use crate::actions::{SelectDown, SelectLeft, SelectRight, SelectUp};
-use crate::scroll::AutoScroll;
 use crate::highlighter::DiagnosticSet;
 #[cfg(not(target_family = "wasm"))]
 use crate::highlighter::LanguageRegistry;
@@ -45,6 +44,7 @@ use crate::input::{
     search::{self, SearchPanel},
 };
 use crate::menu::PopupMenu;
+use crate::scroll::AutoScroll;
 use crate::{Root, history::History};
 
 #[derive(Action, Clone, PartialEq, Eq, Deserialize)]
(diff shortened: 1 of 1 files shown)


huacnlee:
We can't just expose this API, this is a very internal variable.

huacnlee (changes requested):
So, what you original problem? You need find other way.

hlcfan:
I'm working on a feature to restore an editor's previous scroll offset when opens a request https://github.com/hlcfan/beam/pull/6.

I'm thinking either
1. Expose `deferred_scroll_offset` as a public API to set the scroll offset on next paint
2. Or make [update_scroll_offset](https://github.com/longbridge/gpui-component/blob/70d2c44bd62ab16adf3cb791bf776e53d5ea312b/crates/ui/src/input/state.rs#L1650) as public API that can set the offset immediately 

Please lemme know your idea.

crates/ui/src/input/state.rs
```diff
@@ -1877,6 +1877,12 @@ impl InputState {
         self.scroll_handle.offset()
     }
 
+    /// Schedules an editor viewport scroll offset to apply on the next layout pass.
+    pub fn defer_scroll_offset(&mut self, offset: gpui::Point<gpui::Pixels>, cx: &mut Context<Self>) {
```
huacnlee: We can rename to `set_scroll_offset`, so this will match with `scroll_offset`.

The `defer` is a internal thing, for as a API we do not need expose that name.

hlcfan:
Thank you!

PR #2457, opened 2026-06-10
highlighter: prevent empty range style leaks

BeratHundurel:
## Summary

- Ignore empty syntax highlight ranges before style merging
- Filter empty ranges in `unique_styles` so zero-length captures cannot remain active after their position
- Add a focused regression case covering empty ranges leaking styles into later spans

### Before

before

### After

after


madcodelife:
Thank you!

PR #2471, opened 2026-06-16
input: Fix disabled input opacity rendering

huacnlee:
## Summary
- Render disabled inputs by fading background and border colors instead of applying inherited opacity to the whole element tree.
- Fade disabled editor line-height backgrounds used for active lines and inline completion overlays.
- Add a Disabled toggle to the editor example for testing.

Related context: https://github.com/zed-industries/zed/pull/59365

## Test Plan
- git diff --check
- rustfmt --edition 2024 --check crates/story/examples/editor.rs crates/ui/src/input/element.rs crates/ui/src/input/input.rs
- cargo check -p gpui-component-story --example editor

Note: full `cargo fmt --check` currently reports existing formatting diffs outside this change, so this PR validates the touched files directly.

### DIFF
--- a/crates/story/examples/editor.rs
+++ b/crates/story/examples/editor.rs
@@ -77,6 +77,7 @@ pub struct Example {
     soft_wrap: bool,
     show_whitespaces: bool,
     folding: bool,
+    disabled: bool,
     scroll_beyond_last_line: Option<usize>,
     cursor_surrounding_lines: Option<usize>,
     lsp_store: ExampleLspStore,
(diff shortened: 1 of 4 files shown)


hlcfan:
Thanks for checking this issue!

After testing, I found the text color isn't applying the opacity, I made a few more changes to fix this https://github.com/longbridge/gpui-component/commit/995d27bb75d05e0f04d2ca8d94805668e16a213e#diff-2aef75032f4f5a49854fd47def340f61d43e29412aab96912a36e76f016061db.

Applying the opacity at `gpui-component` would work for the issue. My concern would be the one I mentioned in https://github.com/zed-industries/zed/pull/59365, by adding opacity to all sub elements, it adds many tedious changes.

This is the reason I was thinking to expose the `with_element_opacity` and `element_opacity` APIs to public, so that we can simply render the gutter separately without affecting other elements, e.g.
```
window.with_element_opacity(gutter_opacity, |window| {
  // existing logic to paint gutter with opacity
}
```

PR #2477, opened 2026-06-17
input: Add `refresh_lsp` method to re-run LSP state without altering …

### LINKED ISSUE #2476: Allow user to set Lsp providers and refresh the Lsp state after the InputState is created
InputState only refreshes the Editor Lsp on text edits. IE: InputStatue::set_value refreshes the lsp, but it also resets cursor, scroll, and undo history. 

I need to be able to programmatically refresh the Lsp without moving the cursor. 

I was thinking we could add a pub refresh_lsp fn to the InputState impl, which handles this. 

Maybe something like this?

```rust
impl InputState {
    /// Refresh LSP state for the current text without changing the value, selection, or scroll.
    pub fn refresh_lsp(&mut self, window: &mut Window, cx: &mut Context<Self>) {
        self.lsp.update(&self.text, window, cx);
    }
...
```

V/R
Scott

ScottCUSA:
input: Add `refresh_lsp` method to re-run LSP state without altering text or selection

Closes #2476

## Description

Adds a public `InputState::refresh_lsp` method so callers can re-run LSP updates (document colors and semantic tokens) for the current editor text without modifying the value, selection, or scroll position.

As described in [#2476](https://github.com/longbridge/gpui-component/issues/2476), `InputState` only refreshes LSP state as a side effect of text edits. APIs like `set_value` do refresh LSP, but they also reset the cursor, scroll, and undo history. I didn't see another way to refresh LSP decorations after external changes —such as a theme theme switching, provider reconfiguration, or server reconnect—without disturbing the user's editing position.

This PR adds the small, additive API proposed in the issue:

```rust
pub fn refresh_lsp(&mut self, window: &mut Window, cx: &mut Context<Self>) {
    self.lsp.update(&self.text, window, cx);
}
```

## Screenshot

N/A — no visible UI change; this exposes a programmatic API on `InputState`.

## How to Test

1. Build the crate:
   ```bash
   cargo build -p gpui-component
   ```

2. In an app using a code-editor `InputState` with LSP providers configured, call `refresh_lsp` after an external change and confirm document colors / semantic token highlighting refresh without the cursor or selection moving.

## Checklist

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -844,6 +844,13 @@ impl InputState {
         self.disabled = was_disabled;
     }
 
+    /// Refresh Editor LSP state
+    /// 
+    /// This will not trigger any selection changes.
+    pub fn refresh_lsp(&mut self, window: &mut Window, cx: &mut Context<Self>) {
+        self.lsp.update(&self.text, window, cx);
+    }
+
     fn replace_text(
         &mut self,
         text: impl Into<SharedString>,


huacnlee:
That not a good API.

ScottCUSA:
Happy to make revisions. Do you have suggestions?

This is just one of several actions which cannot be triggered directly on the input/editor component by users of the crate. 

For example there's also no public select_all function. Normally dispatching the SelectAll action would do the needful, but components which are not "mounted" in the render tree cannot receive dispatched SelectAll events.

PR #2486, opened 2026-06-19
select: Don't show focus border when appearance is false

dunkmann00:
`Select` now behaves like `Input` does and will not show the focus ring border when `appearance` is `false`.

## How to Test

Observe the `Appearance false with Input` section in the `Select` story. When the select is clicked the focus ring is not shown.

## Checklist

### DIFF
--- a/crates/story/src/stories/select_story.rs
+++ b/crates/story/src/stories/select_story.rs
@@ -1,4 +1,4 @@
-use gpui::*;
+use gpui::{prelude::FluentBuilder as _, *};
 use gpui_component::{button::*, checkbox::*, input::*, select::*, separator::*, *};
 use itertools::Itertools as _;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


huacnlee:
I don't think this modification is correct. We also need to maintain a focus border as much as possible to support keyboard operation.

huacnlee (changes requested):
Try use difference focus style?

dunkmann00:
This doesn't prevent the focus from happening. So you can still use the keyboard to interact with the select. You just don't see the focus ring. I think like in in the story it kind of shows a valid use case, similar to how I came across it in the project I'm working on.

If you want to combine the select with something else and make the whole thing look like one component, it can be helpful to not have the select show its typical ring. But the way I was going to use it is to still use the Select's focus handle to put the ring somewhere else to show it is focused.

dunkmann00:
I made a change to the story to show what I mean as an option of what you could do. Now the `Appearance false with Input` will show a ring around the whole container when you are in the select or input fields. Tabbing with the keyboard shows where you are.

Hopefully I understood your concern correctly.

dunkmann00:
I was playing around with it some more and I do see what you mean. Especially compared to the Input where at least you have the cursor to indicate it is focused. I'll close this for now as I don't really have a good alternative way to style it.

I appreciate the feedback though!

PR #2510, opened 2026-06-24
input: Show the start of a long value after `set_value`

hlcfan:
## Description

Ensure the cursor is at the start when setting text. This fixes the issue when an input value is long, it shows the latter part of the value after setting value.

## How to Test
`test_set_value_move_cursor_to_start`

## Checklist


madcodelife:
Thanks for the fix! 
I tweaked it to keep the caret at the end (like an HTML input) while scrolling the view to the start.

PR #2524, opened 2026-07-01
input: Fix soft-wrap width being applied one frame late

### LINKED ISSUE #2520: Multiline Input soft wrap width incorrect on first frame
## Description

When a multi line Input is first rendered, any text that is soft wrapped will be wrapped at the wrong width. It seems like it will wrap at the previous known, which could be unknown, bigger or smaller than the current width.

## Environment

- GPUI: [1d217ee]
- GPUI Component: [be4c5d3]
- Platform: [mac]

## Steps to Reproduce

1. Set the text of a multi line input that has soft wrap enabled. Make sure lines are long enough to have to soft wrap.
2. If there are no triggers to make it recompute (i.e. no mouse movements), the first time it is rendered it will not wrap at all, since it doesn't have a previous width.
3. Once it is correctly wrapped, it can be triggered again by adjusting the size of the input while it is no longer on screen.
4. Adjust the size of the window and then go back to the view with the input, the wrapping will again be incorrect.

## Screenshots

I recorded this happening with the text area story. Hopefully this demonstrates more clearly what I was trying to describe!

https://github.com/user-attachments/assets/474f6dd1-8f82-4ad0-a1a1-764b0e4498b4

## Expected

Soft wrap to be accurate on the first rendered frame.

## Actual

Text jumps into place after it is already on screen.

dunkmann00:
Reorders some of the layout code in `TextElement::prepaint` to compute the soft wrap width and update the display map so the rest of the layout and painting have the correct soft wrapping in the current frame.

Closes #2520 

## Screenshot

| <video src="https://github.com/user-attachments/assets/474f6dd1-8f82-4ad0-a1a1-764b0e4498b4"></video> | <video src="https://github.com/user-attachments/assets/62c6a66e-91d1-4a6a-83ce-271dd7a71809"></video> |

*Notice in the top textbox the text no longer jumps into place after the mouse is moved.*

## Checklist

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -1562,23 +1562,6 @@ impl Element for TextElement {
         });
 
         let state = self.state.read(cx);
-        let line_height = window.line_height();
-
-        let (visible_range, visible_buffer_lines, visible_top) =
-            self.calculate_visible_range(&state, line_height, bounds.size.height);
-        let visible_start_offset = state.text.line_start_offset(visible_range.start);
-        let visible_end_offset = state
-            .text
-            .line_end_offset(visible_range.end.saturating_sub(1));
-
-        let highlight_styles = self.highlight_lines(
-            &visible_buffer_lines,
-            visible_top,
-            visible_start_offset..visible_end_offset,
-            cx,
-        );
-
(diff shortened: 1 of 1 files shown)


madcodelife:
Thank you!

PR #2533, opened 2026-07-04
input: Expose editor search API

calmondev:
## Description

An application that wants its own search UI on the editor — keeping the editor's matching, highlighting, scrolling and replacing, but drawing the search bar itself — had no supported way to do it: search could only be switched on or off with `.searchable(true)`, and everything else lived inside the built-in `SearchPanel`.

Since this PR was opened, the engine moved into `gpui-base` and its API became public on the editor state (#2691, #2716, #2932, #2953, #2955, #3013): `set_search_query`, `close_search`, `next_search_match`, `previous_search_match`, the replace methods and `search_session()`. This PR, rebased onto that, closes the gaps that still tied a search to the built-in panel and documents the custom-UI path.

### Highlights without the panel

Match highlights were painted only while `search_session.open`, which also shows the panel, so a query set through `set_search_query` counted matches the editor never showed. `SearchSession` now tracks `active` separately from `open`: opening the panel or setting a query activates the session, `close_search` ends it, and the element highlights while it is active. The panel keeps keying its own visibility on `open`. `SearchSession` is `#[non_exhaustive]`, since it gains fields; build it with `Default`.

### The shortcut reaches the application

An editor that is not `searchable` swallowed `Search` and `Replace` instead of propagating them, so a parent view could not bind `Ctrl-F` / `Cmd-F` to its own field. Both handlers now `cx.propagate()` when the editor is not searchable. `searchable` is no longer `#[doc(hidden)]` and is documented for what it gates: the built-in panel and its shortcut, not the search API.

### `current()`

`SearchMatcher::current()` reports the current match as `Option<usize>`, so "no match" is not mistaken for the first one; `current_match_index()` stays.

### Docs

The Editor guide gains a "Custom search UI" section, in English and Chinese: `set_search_query` from the application's own field, navigating and describing the matches, replacing, ending with `close_search`, and taking `Ctrl-F` / `Cmd-F` on the view that owns the field.

## Breaking Changes

None. `SearchSession` is now `#[non_exhaustive]`, so it must be built with `Default` rather than a struct literal; no code in this repository or its examples did otherwise.

## How to Test

- `cargo test -p gpui-base -p gpui-component --lib -- search`: covers the session staying active after `set_search_query` without the panel, the shortcut reaching a host view when the editor is not searchable and opening the panel when it is, and the match label/`current()` semantics.
- `cargo run -p example-editor`: the built-in panel behaves as before, including resuming its last query on reopen.

The rebase, the Base changes and the documentation were written with Claude Code and reviewed by hand; the original API proposal is by @calmondev.

### DIFF
--- a/crates/story/examples/custom_search.rs
+++ b/crates/story/examples/custom_search.rs
@@ -1,0 +1,162 @@
+use gpui::*;
+use gpui_component::{
+    ActiveTheme, Disableable, Icon, IconName, Sizable,
+    button::{Button, ButtonVariants as _},
+    h_flex,
+    highlighter::Language,
+    input::{Input, InputEvent, InputState, Search, TabSize},
+    v_flex,
+};
+use gpui_component_assets::Assets;
+
+pub struct Example {
+    editor: Entity<InputState>,
+    search: Entity<InputState>,
+    _subscriptions: Vec<Subscription>,
+}
+
+impl Example {
+    pub fn new(window: &mut Window, cx: &mut Context<Self>) -> Self {
+        let editor = cx.new(|cx| {
+            InputState::new(window, cx)
+                .code_editor(Language::Rust)
+                .line_number(true)
+                .indent_guides(true)
+                .tab_size(TabSize {
+                    tab_size: 4,
+                    hard_tabs: false,
+                })
+                .soft_wrap(false)
+                .searchable(false)
+                .default_value(include_str!("./fixtures/test.rs"))
+                .placeholder("Enter your code here...")
+        });
+        let search = cx.new(|cx| InputState::new(window, cx).placeholder("Search..."));
+        let _subscriptions = vec![cx.subscribe(
+            &search,
+            |this: &mut Self, search, ev: &InputEvent, cx| match ev {
+                InputEvent::Change => {
+                    let query = search.read(cx).value();
+
+                    this.editor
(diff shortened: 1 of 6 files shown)


huacnlee:
Thanks for the PR, and sorry it sat for so long. I've merged `main` into this branch to resolve the conflicts. Since July the search state moved into `gpui-base` (`InputBaseState::search_session`) and the public API this PR proposed landed there through #2691, #2716, #2932, #2953, #2955 and #3013 (`set_search_query`, `close_search`, `next_search_match`, `previous_search_match`, `replace_current_search_match`, `replace_all_search_matches`, `search_session().matcher.{len, current_match_index, label}`), so the conflicting files take `main`'s version.

Two things this PR wanted were still missing on `main`, and are added in `72310009`:

- Match highlights were painted only while the built-in panel was open, so a query set from a custom UI counted matches the editor never showed. `SearchSession` now tracks `active` separately from `open`.
- An editor that is not `searchable` swallowed `Ctrl-F` / `Cmd-F` instead of propagating it, so a parent could not bind the shortcut to its own field.

`049f7bfd` ports the custom search example to the `examples/` package layout (`cargo run -p example-custom-search`) and the `EditorState` API, and documents the custom search UI in the Editor guide in both languages.

Implementation in these commits was AI-generated and reviewed.

huacnlee:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #2546, opened 2026-07-08
input: Refactor wrap cache to SumTree and skip foreground parsing on large files

zanmato:
## Description

These two changes are related, so I kept them in the same PR but different commits, so it can be reviewed as two.
This is done to improve editing in larger files, but improving this also improves overall editing.
After this change, editing the example json file is faster than Zed (ironically since this uses their sum_tree crate :smile:). Feels on par with Vscode when forcefully enabled highlighting but uses less memory.

### Wrapping refactored to a SumTree

Replaces the `Vec<LineItem>` with a `SumTree<LineItem>`. The vec required O(n) scans for things like row offset, cursor placement or finding the longest line. With SumTree it becomes O(log n) or even O(1) depending on if it's cached. Similar to diagnostics that is already backed by a SumTree. One of the biggest wins is editing though since we no longer need to splice the vec in text_wrapper (`self.lines.splice(rows_range, new_lines);`)

Call sites that previously did `display_map.lines().get(row)` or `.iter().enumerate()` now go through `display_map.line(row)` and `buffer_line_to_display_row(row)`, which resolve via the tree.

### Skip foreground parsing on large files

For large files I've added a threshold for foreground parsing, since we don't even need to attempt it if the file is big enough. It will always time out anyway.. Also attempts to cancel any previous parses, this could be further improved with a parse lock to absolutely prevent parallel parsing.

### Before

https://github.com/user-attachments/assets/58f94c40-a61c-4c8b-8562-fea428e6f17d

### After

https://github.com/user-attachments/assets/dc3f5ce7-b790-4835-8c49-0b514797b185

### Zed comparison

https://github.com/user-attachments/assets/5b7351fd-9f66-497f-a0d7-93eab827842f

This is with the document color provider and linting disabled in the editor. Also folding is disabled because it still is a bit slow for these huge files.

## How to Test

- Generate test file 
```sh
$ (echo '['; for i in $(seq 1 999999); do echo "{\"id\": $i, \"name\": \"user_$i\"},"; done; echo '{"id": 1000000, "name": "user_1000000"}'; echo ']') > big_test.json
```
- Disable/comment lint_document and document_color_provider in editor.rs
```diff
--- a/crates/story/examples/editor.rs
+++ b/crates/story/examples/editor.rs
@@ -713,7 +713,7 @@ impl Example {
             editor.lsp.code_action_providers = vec![lsp_store.clone(), Rc::new(TextConvertor)];
             editor.lsp.hover_provider = Some(lsp_store.clone());
             editor.lsp.definition_provider = Some(lsp_store.clone());
-            editor.lsp.document_color_provider = Some(lsp_store.clone());
+            // editor.lsp.document_color_provider = Some(lsp_store.clone());
 
             editor
         });
@@ -731,7 +731,7 @@ impl Example {
         Self::load_files(tree_state.clone(), PathBuf::from("./"), cx);
 
         let _subscriptions = vec![cx.subscribe(&editor, |this, _editor, _: &InputEvent, cx| {
-            this.lint_document(cx);
+            // this.lint_document(cx);
         })];
 
         Self {
```
- Run editor cargo run --release --example editor
- Disable folding (still a bit slow on these large files)

## Checklist

### DIFF
--- a/crates/ui/src/highlighter/highlighter.rs
+++ b/crates/ui/src/highlighter/highlighter.rs
@@ -470,6 +470,15 @@ impl SyntaxHighlighter {
         self.tree.as_ref()
     }
 
+    /// Apply only the structural `edit` to the existing tree and update the stored text,
+    /// without re-parsing.
+    pub fn edit_tree(&mut self, edit: Option<InputEdit>, text: &Rope) {
+        if let (Some(edit), Some(tree)) = (edit, self.tree.as_mut()) {
+            tree.edit(&edit);
+        }
+        self.text = text.clone();
+    }
+
     /// Returns the language name for this highlighter.
     pub fn language(&self) -> &SharedString {
         &self.language
--- a/crates/ui/src/highlighter/wasm_stub.rs
+++ b/crates/ui/src/highlighter/wasm_stub.rs
@@ -37,6 +37,10 @@ impl SyntaxHighlighter {
         true
     }
 
+    pub fn edit_tree(&mut self, _edit: Option<crate::input::InputEdit>, _text: &ropey::Rope) {
+        // No-op in WASM
+    }
+
     pub fn language(&self) -> &SharedString {
         static EMPTY: SharedString = SharedString::new_static("");
         &EMPTY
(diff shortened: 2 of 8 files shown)


huacnlee:
Thanks for the improve. 

I reviewed and verified it from performance, stability, and also tested syntax highlighting with a 200k-line file. I didn’t find any issues.

Overall, the SumTree-based wrap cache implementation looks complete, and the coordinate
mapping, incremental updates, and large-file parsing path are handled cleanly.

PR #2548, opened 2026-07-08
calendar: added support for custom calendar starting days

### LINKED ISSUE #2517: Override gpui-component localization from the application side
I wanted to localize the calendar component to my language, but that doesn't seem to be possible. There is no available functions nor can I override it using my own translations using the rust-i18n crate. My question is how could I do this without forking the crate?

xrtxn:
Closes #2517 

## Description

Added support for custom starting days for calendar.

## Screenshot

| image | image |

### First day
                DatePickerState::new(window, cx)
                    .date_format("%Y. %m. %d.")
                    .first_day_of_week(Weekday::Mon)

## Checklist

### DIFF
--- a/crates/ui/src/time/calendar.rs
+++ b/crates/ui/src/time/calendar.rs
@@ -1,6 +1,6 @@
 use std::{borrow::Cow, rc::Rc};
 
-use chrono::{Datelike, Local, NaiveDate};
+use chrono::{Datelike, Local, NaiveDate, Weekday};
 use gpui::{
     App, ClickEvent, Context, Div, ElementId, Empty, Entity, EventEmitter, FocusHandle,
     InteractiveElement, IntoElement, ParentElement, Render, RenderOnce, SharedString, Stateful,
@@ -266,6 +266,8 @@ pub struct Calendar {
     style: StyleRefinement,
     /// Number of the months view to show.
     number_of_months: usize,
+    /// The first day of the week. Defaults to Sunday.
+    first_day_of_week: Weekday,
 }
 
 /// Use to store the state of the calendar.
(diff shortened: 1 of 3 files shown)


huacnlee:
Hi, `weekday_names`, `month_names` is not a good solution. I have make change of rust-i18n to support this feature and make PR in #2563

xrtxn:
Thanks, this should work for me, this was my original idea too, but unsure how to implement it. However, please check out .first_day_of_week() as well, as this is still missing. If you rather I can implement the changes needed for this updated version,

huacnlee:
first_day_of_week is ok to merge, you can cleanup you PR, then I will merge it.

xrtxn:
Please check it out, I cleaned up the code

huacnlee (commented):
Thanks for this! The grid math for the start day is correct, and it stays backward-compatible for the default (Sunday).

One blocking issue: the weekday header row isn't reordered, so any `first_day_of_week` other than `Sun` makes the whole calendar misaligned.

In `render_days` (calendar.rs) the header is a fixed Sunday-first list and is rendered in that order regardless of the start day:

```rust
let weeks = [t!("Calendar.week.0") /*Sun*/, ..., t!("Calendar.week.6") /*Sat*/];
// weeks.iter().map(render_week)  // always Sun -> Sat
```

But `days()` now shifts the date columns to start on `first_day_of_week`. So with `first_day = Mon`, column 0 shows the "Sun" header on top of Monday's dates.

Example - Feb 2023 (the 1st is a Wednesday), `first_day = Mon`:
- date columns become `Mon Tue Wed Thu Fri Sat Sun`
- header still renders `Sun Mon Tue ...` -> off by one

Fix: rotate the `weeks` array by `self.first_day_of_week.num_days_from_sunday()` before rendering, using the same offset as the grid.

- `test_days` still only checks `Weekday::Sun`. Please add a `Weekday::Mon` case so the offset is locked in.

xrtxn:
Thanks for the review, I didn't spot this, because I set the the day.0 to Monday, which is admittedly not the right option. I updated it, now it should hopefully work properly. But this was the original reason I renamed the i18n `day.0` key to a proper `sunday` because in different countries the first day is different.

Regardless, it should be fixed now, I added a test, please check it out again.

huacnlee:
Update your PR title and description please. Then I will merge this.

xrtxn:
Fixed it.

PR #2549, opened 2026-07-08
editor: Add to option to keep completion menu at original position.

palaska:
## Description

The completion popover recomputes its position from the live caret bounds on every render, so it slides right as you type the query, reading as jittery.

This pins the popover to where the completion session opened and keeps it there until the session ends (`hide`) or a new one begins, matching the behavior of editors like VS Code and Zed.

## Screenshot

**Before**
sticky suggestion before

**After**
sticky suggestion after

## Break Changes

None

## How to Test

1. `cargo run --example editor`
2. In the editor, type a word that triggers the completion popover (e.g. start typing so the menu appears).
3. Keep typing to extend the word.

**Before:** the popover slides right, tracking the caret.
**After:** the popover stays pinned where the completion opened.

Dismissing (Esc) and re-triggering elsewhere re-pins it at the new location.

## Checklist

### DIFF
--- a/crates/ui/src/input/popovers/completion_menu.rs
+++ b/crates/ui/src/input/popovers/completion_menu.rs
@@ -177,6 +177,9 @@ pub struct CompletionMenu {
     /// The offset of the first character that triggered the completion.
     pub(crate) trigger_start_offset: Option<usize>,
     query: SharedString,
+    /// Screen position the popover is pinned to for a completion session.
+    /// Cleared on `hide` and when a new session starts.
+    anchor: Option<Point<Pixels>>,
     _subscriptions: Vec<Subscription>,
 }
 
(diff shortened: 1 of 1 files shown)


ScottCUSA:
VS Code and Zed's completion popovers follow the cursor like your Before example.

Not to say gpui-component's couldn't be better, the other editor's auto-close the popover after you finish a word, or there are no matches, then re-open when you start typing a second word.

palaska:
> Zed's completion popover follows the cursor like your Before example.
> 
> Though maybe a little closer, and it auto-closes after you hit space to finish a word, then re-opens when you start typing a second word.

Hmm thats true actually. Yeah it felt jittery and laggy probably because as you said it doesn't follow the cursor as closely as the others so the after behavior felt much nicer to me.

If you prefer keeping the old behavior, I guess I can make this gated by a config. Feel free to close this if you don't think it's necessary :)

huacnlee:
I'd like this feature, but I'd like this to as an option and default to before version.

PR #2557, opened 2026-07-11
highlighter: Highlight fenced code blocks in Markdown source

BeratHundurel:
## Description

Markdown fenced code blocks were rendered with the generic code-block color because captured fence language names were skipped by the injection pipeline.

This change resolves registered fence languages and applies their tree-sitter highlighting above the host Markdown styling. To keep performance and memory usage bounded, highlight queries are created lazily, shared between blocks using the same language, and reused across incremental parses. Unknown or oversized language identifiers are ignored, while the existing layer, range, byte, and parsing-time limits remain in effect.

Tests cover language aliases, query reuse, style precedence, unknown languages, and existing Markdown injection limits.

AI-assisted implementation was reviewed and validated.

## Screenshot
image

## Testing

- `cargo check -p gpui-component --features tree-sitter-languages`


huacnlee:
Thank you.

PR #2559, opened 2026-07-13
setting: Add `title_suffix` for setting page header

madcodelife:


crates/story/src/stories/settings_story.rs
```diff
@@ -139,6 +139,15 @@ impl SettingsStory {
                 .resettable(resettable)
                 .default_open(true)
                 .icon(Icon::new(IconName::Settings2))
+                .title_suffix(|_, _| {
```
huacnlee, line 142: Rename to toolbar

huacnlee (dismissed):

PR #2569, opened 2026-07-14
input: add Lsp::show_document hook following LSP window/showDocument

ylinwind:
## What

Adds an optional `Lsp::on_show_document` host hook on `InputState`, called at the start of `go_to_definition` before the built-in handling. The hook is modeled after the LSP [`window/showDocument`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument) request:

```rust
pub type ShowDocumentHandler =
    Rc<dyn Fn(&lsp_types::ShowDocumentParams, &mut Window, &mut App) -> bool>;
```

The component builds `ShowDocumentParams` from the resolved `LocationLink` (`uri`, `external` for http(s) targets, `take_focus`, `selection` from `targetSelectionRange`).

- Return `true`: the host has shown the document; the component does nothing further.
- Return `false` (or hook unset): falls through to the existing behavior — `external` URIs open in the browser, anything else jumps within the current document.

The default in-document jump now selects `targetSelectionRange` (the range the spec says should be selected and revealed when following a link) instead of `targetRange`.

## Why

The built-in Go to Definition can only jump within the current document or open a browser URL. Language providers that resolve symbols into virtual/external documents (e.g. NaviScript stdlib symbols resolving to `navi-doc://module/name`) need a way to route Cmd+Click to an in-app docs window. Without a hook, non-http cross-file locations would be misinterpreted as same-document ranges and jump to a wrong position.

`window/showDocument` is the spec's standard "client, show this document" abstraction, so the hook reuses its params shape rather than inventing a new one.

## Notes

- Branch is based on `d1c48828` (the rev currently pinned by longbridge-gpui's Cargo.lock) so it can be consumed immediately; happy to rebase onto `main` if preferred.
- `go_to_definition` (crate-private) gains a `&mut Window` parameter; both internal call sites already had a window in scope.

### DIFF
--- a/crates/ui/src/input/lsp/mod.rs
+++ b/crates/ui/src/input/lsp/mod.rs
@@ -19,6 +19,14 @@ pub use document_colors::*;
 pub use hover::*;
 pub use semantic_tokens::*;
 
+/// Host hook to intercept opening an LSP location (Go to Definition).
+///
+/// Called before the built-in behavior. Return `true` if the host handled
+/// the location (e.g. opened a docs window for a virtual/external URI);
+/// return `false` to fall through to the default handling (http(s) URIs
+/// open in the browser, anything else jumps within the current document).
+pub type OpenLocationHandler = Rc<dyn Fn(&lsp_types::LocationLink, &mut Window, &mut App) -> bool>;
+
 /// LSP ServerCapabilities
 ///
 /// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities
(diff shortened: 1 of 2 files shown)


crates/ui/src/input/lsp/mod.rs
```diff
@@ -35,6 +43,9 @@ pub struct Lsp {
     pub document_color_provider: Option<Rc<dyn DocumentColorProvider>>,
     /// The range semantic tokens provider.
     pub semantic_tokens_provider: Option<Rc<dyn DocumentRangeSemanticTokensProvider>>,
+    /// Optional host hook to intercept Go to Definition locations
+    /// (see [`OpenLocationHandler`]).
+    pub on_open_location: Option<OpenLocationHandler>,
```
huacnlee: Requires a reference from ttps://microsoft.github.io/language-server-protocol/specifications/lsp/

ylinwind: I'll handle this myself — reworked in 3c1bd602 to follow the spec's `window/showDocument` request (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument): the hook is renamed to `on_show_document` and now takes `lsp_types::ShowDocumentParams` instead of a raw `LocationLink`. The default in-document jump also selects `targetSelectionRange` now, per the spec.

Issue #2325, opened 2026-04-30
Code editor styles broken on web version


tristanpoland:
Coloring seems to be broken on the web version of the code editor 

<img width="1408" height="843" alt="Image" src="https://github.com/user-attachments/assets/81c0129b-6dc0-44df-8035-06190bfa2aa6" />

huacnlee:
For the Wasm file size reason, I have not include the tree-sitter into Wasm build.

tristanpoland:
> For the Wasm file size reason, I have not include the tree-sitter into Wasm build.

Oh interesting, I suppose that makes sense! I wonder if it would be worth packaging tree sitter as a second file to load asynchronously or maybe even optionally.

I haven't done much with WASM yet so no idea how that would work.

PR #2570, opened 2026-07-15
highlighter: Preserve text color without syntax highlighting

MuNeNiCK:
Related to #2325

## Description

When the `tree-sitter` feature is disabled, the stub implementation of `SyntaxHighlighter::styles` returns an empty style list.

The input renderer builds text runs from that list. When the list is empty, no colored text run is created, so GPUI paints the text with its default black instead of preserving the input's text color.

The native highlighter returns the requested range with `HighlightStyle::default()` when no styles match. This change aligns the stub implementation with that behavior so the input text color is preserved when syntax highlighting is disabled.

## How to Test

- Run `cargo check -p gpui-component --target wasm32-unknown-unknown --lib`.
- Build a web application that uses gpui-component with `default-features = false`.
- Enter text in a CodeEditor with a dark theme.
- Verify that existing and newly entered text use the input's text color.

## Checklist

- [N/A] Passed `cargo run` for story tests related to the changes.
- [N/A] Tested macOS, Windows and Linux platforms performance (not platform-specific).

### DIFF
--- a/crates/ui/src/highlighter/wasm_stub.rs
+++ b/crates/ui/src/highlighter/wasm_stub.rs
@@ -21,10 +21,11 @@ impl SyntaxHighlighter {
 
     pub fn styles(
         &self,
-        _range: &Range<usize>,
+        range: &Range<usize>,
         _theme: &HighlightTheme,
     ) -> Vec<(Range<usize>, HighlightStyle)> {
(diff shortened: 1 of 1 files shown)


MuNeNiCK:
Thanks for merging the PR. I really appreciate it.

PR #2572, opened 2026-07-15
input: Avoid painting unset gutter background

MuNeNiCK:
Related to #2411

## Description

The line-number gutter is painted as a separate quad on top of the editor background.

When `editor.gutter.background` is unset, the current implementation falls back to `editor_background()`. Since the editor already paints that background, translucent colors are composited twice in the gutter and make it appear darker than the text area.

The fallback can also conflict with a background applied through the editor's `Styled` implementation because the gutter still paints the theme's editor background separately.

This change paints the gutter background only when `editor.gutter.background` is explicitly configured. When it is unset, the editor background remains visible without an additional paint layer. Themes that define a dedicated gutter background continue to render it as before.

## Screenshots
Before:
editor-gutter-before

After:
editor-gutter-after

## How to Test

- Run `cargo check -p gpui-component`.
- Run `cargo run` and open the Editor story.
- Use a theme with a translucent `editor.background` and no `editor.gutter.background`.
- Verify that the gutter and text area have the same visual background.
- Set `editor.gutter.background` and verify that the explicit gutter background is still rendered.

## Checklist

- [N/A] Tested macOS, Windows and Linux platforms performance (not platform-specific).

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -2122,25 +2122,18 @@ impl Element for TextElement {
         if let Some(line_numbers) = prepaint.line_numbers.as_ref() {
             offset_y += invisible_top_padding;
 
-            // Gutter background: prefer the dedicated `editor.gutter.background`
-            // theme key, falling back to the editor background so existing
-            // themes render unchanged.
-            let gutter_bg = cx
-                .theme()
-                .highlight_theme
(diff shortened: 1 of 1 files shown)


MuNeNiCK:
Thanks for merging the PR. I really appreciate it.

PR #2574, opened 2026-07-17
select, combobox: Use popover background for dropdown menus

sh4den:
Select and Combobox paint their dropdown menu panel with `tokens.background`,
while PopupMenu and Popover use `tokens.popover`. With a theme that has a
translucent background color (transparent window setups), open select menus
become see-through and hard to read while every other floating panel stays
opaque.

This paints both menus with `tokens.popover` and its paired
`popover_foreground`, the same pair `popover_style()` uses, so they match the
rest of the popups. In the default themes `popover` falls back to
`background`, so nothing changes visually there. No API changes.

## Screenshot

|   image | image |

## How to Test

Run `cargo run` and open the Select and Combobox stories, menus should look
unchanged with the default themes (`background` and `popover` are the same
color there). To see the fix, set the theme's `background` to a translucent
color and open a select menu: it stays opaque instead of showing the content
behind it.

## Checklist

### DIFF
--- a/crates/ui/src/combobox.rs
+++ b/crates/ui/src/combobox.rs
@@ -986,7 +986,7 @@ fn render_popup_shell<D: SearchableListDelegate + 'static>(
                     v_flex()
                         .occlude()
                         .mt_1p5()
-                        .bg(cx.theme().tokens.background)
+                        .bg(cx.theme().tokens.popover)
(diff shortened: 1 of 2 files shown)


madcodelife:
Thanks.

PR #2576, opened 2026-07-18
combobox: Add value-based selection API

### LINKED ISSUE #2575: Add value-based selection API to ComboboxState
## Description

`ComboboxState` exposes `selected_values()` for reading the current selection, but it does not provide the inverse operation for selecting items by their values. Callers can only use `set_selected_indices`, even though the current delegate already knows how to resolve a value through `SearchableListDelegate::position`.

`SelectState` already provides the corresponding `set_selected_value` API. Adding a value-based setter to `ComboboxState` would make programmatic selection symmetric and allow it to always use the component's current delegate, including after `set_items` replaces the item catalog.

One possible API is:

```rust
pub fn set_selected_values(
    &mut self,
    values: &[<D::Item as SearchableListItem>::Value],
    window: &mut Window,
    cx: &mut Context<Self>,
)
```

The exact argument shape could follow the project's preferred iterator conventions. Like `set_selected_indices`, this should update the committed selection and snapshot without emitting a user-originated change event.

## Environment

- GPUI: Git dependency used by the current `main` branch
- GPUI Component: v0.5.2 / current `main`
- Platform: All

## Steps to Reproduce

1. Create a `ComboboxState<D>` with a searchable-list delegate.
2. Keep one or more `D::Item::Value` values that should be selected programmatically.
3. Try to apply those values to the existing `ComboboxState`.
4. Observe that the public API only accepts `IndexPath` values through `set_selected_indices`; the delegate needed to resolve values is internal.

## Screenshots

Not applicable.

## Expected

`ComboboxState` provides a value-based selection method, analogous to `SelectState::set_selected_value`, which resolves values using its current delegate and updates the selection.

## Actual

Callers must retain a separate copy of the delegate or item catalog, resolve values to indices themselves, and call `set_selected_indices`. That duplicates state outside the component and can become stale after the component's items are replaced.

suxiaoshao:
Closes #2575

## Description

Adds `ComboboxState::set_selected_values` to resolve values through the current delegate and replace the committed selection without emitting user-originated events. This keeps programmatic selection aligned after `set_items`, synchronizes the adapter snapshot, ignores unknown values, and preserves caller order.

Includes focused tests for current delegate resolution, missing and empty values, snapshot synchronization, and event behavior, plus English and Simplified Chinese documentation.

## How to Test

- `cargo check -p gpui-component --tests`
- `cargo clippy -p gpui-component --all-targets -- --deny warnings`
- `rustfmt --edition 2024 --check crates/ui/src/combobox.rs`
- `git diff --check`

## Checklist

- [x] Reviewed AI-assisted changes for correctness.
- [N/A] Story testing; no visual behavior changed.
- [N/A] Platform testing; the API is platform-independent.

### DIFF
--- a/crates/ui/src/combobox.rs
+++ b/crates/ui/src/combobox.rs
@@ -289,6 +289,30 @@ where
         self.state.selection()
     }
 
+    /// Replace the entire selection set by item values.
+    ///
+    /// Values are resolved through the current delegate. Values that cannot be resolved are
+    /// ignored. This updates the committed selection and snapshot without emitting a
+    /// [`ComboboxEvent`].
+    pub fn set_selected_values(
+        &mut self,
+        values: &[<D::Item as SearchableListItem>::Value],
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        let selected_indices = {
+            let list = self.state.list.read(cx);
+            let delegate = &list.delegate().delegate;
+
+            values
+                .iter()
+                .filter_map(|value| delegate.position(value))
+                .collect::<Vec<_>>()
+        };
+
+        self.set_selected_indices(selected_indices, window, cx);
+    }
+
     /// Replace the entire selection set.
     pub fn set_selected_indices(
         &mut self,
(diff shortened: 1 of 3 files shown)


madcodelife:
Thanks.

PR #2578, opened 2026-07-20
tab: Add max_width to TabBar

ScottCUSA:
## Description

Adds `max_tab_width` to `TabBar`, capping each tab rendered width. Labels set via `.label()` truncate with an ellipsis past the limit; icon-only tabs are unaffected. Tabs with `.child()` content get the width constraint but handle their own truncation. The overflow menu still shows full labels.

## Screenshot

| Before | After |
| ------ | ----- |
| image | image |

## How to Test

1. `cargo run` and open the Tabs story.
2. Toggle “Max tab width”; the enabled value is fixed at 110px.
3. Confirm long labels truncate, short labels keep their intrinsic width, custom children remain constrained, and the overflow menu shows full text.
4. Run `cargo test -p gpui-component tab::`.

## Checklist

- [x] Read CONTRIBUTING and followed the guidelines.
- [x] Reviewed changes, confirmed AI-generated code is accurate.
- [x] Passed Story compilation and tab tests.
- [ ] Tested macOS, Windows, Linux performance (if platform-specific).

### DIFF
--- a/crates/ui/src/tab/tab.rs
+++ b/crates/ui/src/tab/tab.rs
@@ -750,13 +750,6 @@ impl RenderOnce for Tab {
             inner_content.into_any_element()
         };
 
-        // Under `max_width` the label is the only part that gives way, so
-        // hold the prefix and suffix (e.g. a close button) at their full size.
-        let fixed = move |element: AnyElement| match max_width {
-            Some(_) => div().flex_shrink_0().child(element).into_any_element(),
-            None => element,
-        };
-
         self.base
             .id(self.ix)
             .selected(self.selected)
@@ -849,9 +842,23 @@ impl RenderOnce for Tab {
                         ),
                 )
             })
-            .when_some(self.prefix, |this, prefix| this.child(fixed(prefix)))
+            // Under `max_width` the label is the only part that gives way, so
+            // hold the prefix and suffix (e.g. a close button) at their full size.
+            .when_some(self.prefix, |this, prefix| {
+                this.child(
+                    div()
+                        .when_some(max_width, |this, _| this.flex_shrink_0())
+                        .child(prefix),
+                )
+            })
             .child(inner_element)
-            .when_some(self.suffix, |this, suffix| this.child(fixed(suffix)))
+            .when_some(self.suffix, |this, suffix| {
+                this.child(
+                    div()
+                        .when_some(max_width, |this, _| this.flex_shrink_0())
+                        .child(suffix),
+                )
+            })
             .when_some(self.on_click.clone(), |this, on_click| {
                 this.on_click(move |event, window, cx| on_click(event, window, cx))
             })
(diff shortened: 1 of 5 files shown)


huacnlee:
Thanks for the update. I still think the current implementation is too complex for this requirement. It introduces manual text measurement and several special layout branches for labels, prefixes, and suffixes, which makes the core `Tab` rendering logic harder to maintain.

I also tried simplifying it to use the native `max_w`/flex layout directly, but that attempt did not correctly constrain the outer tab bounds in all variants, so I have reverted it.

Could you please explore another approach with a smaller, more focused diff? Ideally, the solution should rely on GPUI layout primitives without manually measuring text or restructuring unrelated parts of the existing tab layout. Please verify that:

- tabs shorter than the limit keep exactly their original width;
- the outer bounds of longer tabs do not exceed `max_tab_width`;
- truncation works across the existing tab variants; and
- prefixes and suffixes continue to behave correctly.

Please keep follow-up changes limited to what is necessary for `max_tab_width`.

huacnlee (changes requested):
Please try to make it simple.

crates/ui/src/tab/tab.rs
```diff
@@ -564,6 +566,46 @@ impl Tab {
         self.tab_bar_prefix = Some(tab_bar_prefix);
         self
     }
+
+    pub(super) fn max_tab_width(mut self, max_tab_width: Option<Pixels>) -> Self {
```
huacnlee: Here is in `Tab`, so we can rename to `max_width`.

ScottCUSA:
> Thanks for the update. I still think the current implementation is too complex for this requirement. It introduces manual text measurement and several special layout branches for labels, prefixes, and suffixes, which makes the core `Tab` rendering logic harder to maintain.
> 
> I also tried simplifying it to use the native `max_w`/flex layout directly, but that attempt did not correctly constrain the outer tab bounds in all variants, so I have reverted it.
> 
> Could you please explore another approach with a smaller, more focused diff? Ideally, the solution should rely on GPUI layout primitives without manually measuring text or restructuring unrelated parts of the existing tab layout. Please verify that:
> 
> * tabs shorter than the limit keep exactly their original width;
> * the outer bounds of longer tabs do not exceed `max_tab_width`;
> * truncation works across the existing tab variants; and
> * prefixes and suffixes continue to behave correctly.
> 
> Please keep follow-up changes limited to what is necessary for `max_tab_width`.

There was a bug in the gpui text layout pipeline which caused the need for the text measurement. That has been fixed, and I dropped the fn, but it requires bumping the gpui version. See https://github.com/zed-industries/zed/pull/61107 

Also, I have made the reqested changes from max_tab_width to max_width. Let me know if you'd like additional changes.

crates/ui/src/tab/tab.rs
ScottCUSA: done

ScottCUSA:
@huacnlee It seems your auto-merge did not execute while this could be applied without conflicts because of your open review requesting changes. Attempting to rebase of the latest.

ScottCUSA:
Updated Tab Story: 

image

PR #2586, opened 2026-07-22
input: Fix prevent single-line input panic on multiline text (#2552)

### LINKED ISSUE #2552: Editor component panics when search is initiated while multiple lines is selected
## Description

Editor component panics when search is initiated while multiple lines is selected.

## Environment

- GPUI Component: main branch
- Platform: macOS 26

## Steps to Reproduce

Test case in my fork - https://github.com/gintsgints/gpui-component/blob/673a741bc3543f4a187a3b019d8a64c22ab2d450/crates/ui/src/input/search.rs#L646

## Expected

Should discuss. Eather strip search to one line or search should expand to multiline.

## Actual

Test fails with - "text argument should not contain newlines"

chulingera2025:
## Summary

Fixes #2552.

Activating find (cmd-f) while a selection spans multiple lines panicked the app with:

```
text argument should not contain newlines
```

The multiline selection was pushed into the **single-line** search input via `set_value`, and painting a single line whose text contains `\n` panics in gpui's text shaper (`text_system.rs`).

## Changes

- **Central fix** in `InputState::normalize_input`: for single-line inputs, collapse newlines to spaces (matching browser search/text inputs) — `\r\n`/`\n`/`\r` become a single space. This covers *every* edit path (`set_value`, `insert`, `paste`, IME), so multiline text stays usable instead of crashing the input.
- Removed the now-redundant manual `\n` strip in `paste` (the central normalization handles it, including `\r`).
- Search populated from a multiline selection becomes a single-line, space-joined query.
- Added tests:
  - `test_single_line_collapses_newlines_to_spaces` — `set_value`/`insert` with mixed `\n`, `\r\n`, `\r` all collapse to spaces.
  - `test_search_with_multiline_selection` — regression test for the original panic; asserts the single-line search input never holds a newline.

## Verification

Full input test suite passes (90 tests) and clippy is clean.

### DIFF
--- a/crates/macros/src/derive_into_plot.rs
+++ b/crates/macros/src/derive_into_plot.rs
@@ -84,8 +84,17 @@ pub fn derive_into_plot(input: TokenStream) -> TokenStream {
                 let hitbox = window.insert_hitbox(bounds, gpui::HitboxBehavior::Normal);
 
                 let overlay = (|| {
-                    // Read the cursor position recorded by the previous frame's mouse handler.
-                    let position = Self::__plot_tooltip_cursor(global_id, window).get()?;
+                    // The cell only gates visibility: it holds the cursor recorded by the
+                    // mouse handler / per-frame sync in `paint`, which is one frame stale
+                    // during scrolling. Rendering that cached point while the bounds move
+                    // makes the tooltip jitter, so derive the position from the live mouse
+                    // position and this frame's bounds instead.
+                    Self::__plot_tooltip_cursor(global_id, window).get()?;
+                    let mouse = window.mouse_position();
+                    if !bounds.contains(&mouse) {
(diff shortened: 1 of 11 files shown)


huacnlee (changes requested):
That not a good solution.

We need support multiple paste. For example in Chrome if we have multiline text, the search input will replace `\n` to ` ` space.

chulingera2025:
> That not a good solution.
> 
> We need support multiple paste. For example in Chrome if we have multiline text, the search input will replace `\n` to ` ` space.

I get it now, I'll fix it.

chulingera2025:
The real machine test passed, the function is normal
@huacnlee

huacnlee (approved):

Issue #1621, opened 2025-11-17
Lagging UI when running examples via cargo run on gpui_component 0.4.0


J2ane:
**Environment**
- Kernel: 6.16.8, x86_64
- Desktop Environment: XFCE
- Display Server: X11
- gpui version: 0.2.2
- gpui_component version: 0.4.0
- Rust version: rustc 1.91.0 (2025-10-28)

**Problem Description**
When running the example of gpui_component 0.4.0 with cargo run, the application becomes noticeably laggy:
- All elements that contain scrollbars — such as List, ScrollView, Table, and scrollable Panels — become extremely laggy when scrolling.
- CPU hits 100% on a single thread
- General UI interactions also feel slightly laggy
- Maximizing or restoring the window also causes noticeable stuttering.

However, the issue does NOT occur when:
- Using gpui_component 0.3.1 (both cargo run and cargo build)
- Running 0.4.0 from a binary built using cargo build (non-run mode)

So the issue happens only under gpui_component 0.4.0 + cargo run.

<img width="2879" height="1770" alt="Image" src="https://github.com/user-attachments/assets/af07b25b-712e-4bcf-97de-d2eaaf1ef057" />

han1548772930:
https://github.com/longbridge/gpui-component/issues/856

huacnlee:
Have you tried to run as `cargo run --release`, and how is the performance?

J2ane:
@huacnlee @han1548772930 
When running with cargo run --release or cargo build --release:
- In the main window, when not on the Table page, scrolling in List and other Scrollable elements works at normal speed.
- On the Table page, scrolling in List and Scrollable elements inside the main window still experiences a 1–2 second delay (no delay in version 0.3.1).
- Resizing or maximizing/restoring the window works at normal speed.
- Adjusting the structure width in a Resizable window still has a ~1 second delay (no delay in version 0.3.1).
- Dragging panels in Tiles still has approximately a 1-second delay (no delay in version 0.3.1).

zanmato:
Can you try running `git bisect` from 0.3.1 and see which commit introduced the performance regression?

J2ane:
@zanmato 

```
git bisect good 
二分查找中：在此之后，还剩 0 个版本待测试 （大概 0 步） 
2f3623f70afcea6ac0f3e41cd2444d64da3adfcf plot: Rename Label and Axis include Plot prefix (#1597)
```

------

[...]

PR #2592, opened 2026-07-24
chore: Provide IBM Plex Sans to fix story app lag to nix

Ziqi-Yang:
## Summary

- add IBM Plex Sans to the Nix development environment
- provide a deterministic Fontconfig configuration when running the story app with Cargo
- apply the same font configuration to the packaged executable wrapper

## Why

GPUI resolves the Linux system UI font to IBM Plex Sans. On NixOS, that font was not available inside the development shell, so text rendering repeatedly took the missing-font fallback path. With `RUST_BACKTRACE=1` enabled by the shell, those fallback errors also captured backtraces and caused high single-thread CPU usage during scrolling and other UI interactions.

Providing the expected font removes that hot path. In an 8-second runtime sample, CPU time dropped from approximately 4.05 seconds to 0.97 seconds, and profiling no longer showed unwind/backtrace work.

## Testing

- `nix flake check --no-build`
- `nix develop --command cargo run --release`
- verified `fc-match "IBM Plex Sans"` resolves to `IBMPlexSans-Regular.otf` inside the dev shell
- `nix build .#defaultPackage.x86_64-linux`
- launched the packaged executable through its generated wrapper

Related issue: #1621

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -14,6 +14,12 @@
         pkgs = import nixpkgs {
           inherit system overlays;
         };
+        uiFont = pkgs.ibm-plex.override {
+          families = [ "sans" ];
+        };
+        fontsConf = pkgs.makeFontsConf {
+          fontDirectories = [ uiFont ];
+        };
         build-dependencies = with pkgs; [
           pkg-config # For dynamically linked libraries
           makeWrapper # To provide LD_LIBRARY_PATH to the final binary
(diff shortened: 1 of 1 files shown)


Ziqi-Yang:
(Human) I've tested the PR. Before PR, scrolling in the gallary application (`cargo run --release`) feels very laggy. This PR fixes the issue for me.

PR #2598, opened 2026-07-26
feat(kbd): add keystroke recorder

Ziqi-Yang:
Add a reusable KeystrokeRecorder beside Kbd. It captures canonical GPUI keybinding sequences, renders captured keys with Kbd, uses component Buttons for record/stop/clear controls, supports configurable sequence length, and emits Changed or Cancel events.

The Kbd story includes the recorder and locale strings cover all existing component locales.

Tested with:
- cargo test -p gpui-component --lib (320 passed)
- gpui-component-story check reaches the native dependency build, then is blocked in this environment by missing glib/gobject pkg-config files

### DIFF
--- a/crates/story/src/stories/kbd_story.rs
+++ b/crates/story/src/stories/kbd_story.rs
@@ -3,12 +3,17 @@ use gpui::{
     Styled, Window,
 };
 
-use gpui_component::{h_flex, kbd::Kbd, v_flex};
+use gpui_component::{
+    h_flex,
+    kbd::{Kbd, KeystrokeRecorder},
+    v_flex,
+};
 
 use crate::section;
 
 pub struct KbdStory {
     focus_handle: gpui::FocusHandle,
+    recorder: Entity<KeystrokeRecorder>,
 }
 
 impl super::Story for KbdStory {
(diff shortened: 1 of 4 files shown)


huacnlee (changes requested):
Thanks for the thorough work here, and for covering the story and all the locales — the feature itself looks complete and works.

I'm going to hold off on merging this one, and I want to be upfront about why so it's not wasted effort.

The core reason is a design bar, not a code problem: a new component in the library needs a general-purpose API and a neutral visual base that fits any app. Right now the default styling is fairly opinionated — the recording indicator especially reads as a specific product choice rather than a building block. That makes it hard to drop into arbitrary UIs without fighting the style. For reference, even Shadcn doesn't ship a component like this, which is part of why I want a higher bar before we take one on.

Where I could see this landing: if the recorder looked like a plain macOS-native control — no special/branded elements like the recording indicator — so it reads as a neutral primitive that any app can theme. If you're up for reshaping it in that direction, I'm happy to reconsider.

Again, appreciate the effort — this is a "design fit" call, not a quality one.

Ziqi-Yang:
Thanks for reviewing it! Sorry and I didn't realize it that codex automaticaly send these PRs out. Some are possibly not stuiable here yet. I'd close these PRs for now. If you think some of them are worth a merge, then reopen it and I'd like to provide assistant. :)

PR #2599, opened 2026-07-26
feat(input): support application text highlights

Ziqi-Yang:
Add InputState set_text_highlights and text_highlights APIs for application-owned UTF-8 byte ranges. Presentation highlights now work in plain, auto-growing, and code-editor inputs and compose with existing syntax, semantic-token, hover, and diagnostic styles.

Ranges are clamped on assignment, visible-range clipping is handled during layout, and plain inputs receive default runs for unstyled gaps.

Tested with:
- cargo test -p gpui-component --lib (320 passed)

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -35,20 +35,24 @@ fn compose_text_highlights(
     text_highlights: &[(Range<usize>, HighlightStyle)],
     visible_byte_range: Range<usize>,
 ) -> Option<Vec<(Range<usize>, HighlightStyle)>> {
-    let text_highlights = text_highlights.iter().filter_map(|(range, style)| {
+    let has_visible_highlight = text_highlights.iter().any(|(range, _)| {
+        range.start < visible_byte_range.end && range.end > visible_byte_range.start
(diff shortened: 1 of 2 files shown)


huacnlee:
I reviewed and narrowed this change to keep the diff as small as possible while preserving the original input implementation.

The current revision keeps the existing `highlight_lines` flow and only adds the necessary composition points for application-provided text highlights. It also clips ranges to valid UTF-8 boundaries, clears stale highlights after text mutations, avoids applying source byte ranges to masked display text, and adds focused regression coverage. The existing IME underline behavior remains untouched.

For follow-up changes, please make them only when required and keep them narrowly scoped to the PR goal. Avoid refactoring or changing unrelated input core logic as part of this feature.

huacnlee:
Closing this PR in favor of #2602.

The replacement was reimplemented from the latest `main` branch with a collection-based Decorations API. This keeps decoration ownership isolated across multiple producers, follows GPUI/GPUI Component naming and styling conventions, references Monaco's Decorations Collection design in the key API docs, and includes an interactive example and focused regression coverage.

PR #2601, opened 2026-07-26
input: Fix IME composition rendering underline

nuskey8:
## Description

IME marked text was not underlined when the input already contained text and syntax highlighting was disabled. With syntax highlighting enabled, the underline was only applied when a complete highlight range was contained in the marked range. As a result, the composition underline could disappear or become incomplete when it crossed Markdown or other syntax-highlight boundaries.

This PR fixes the issue mentioned above.

## How to Test

This can be reproduced with the following code.

```rust
use gpui::*;
use gpui_component::{
    input::{Input, InputState},
    *,
};
use gpui_component_assets::Assets;

struct Example {
    editor: Entity<InputState>,
}

impl Example {
        let editor = cx.new(|cx| {
            InputState::new(window, cx)
                .code_editor("markdown")
                .default_value("# Existing Markdown\n\nCompose here: ")
        });
        Self { editor }
    }
}

impl Render for Example {
    fn render(&mut self, _: &mut Window, _: &mut Context<Self>) -> impl IntoElement {
        v_flex()
            .size_full()
            .p_5()
            .gap_4()
            .child(Input::new(&self.editor).h(px(200.)))
    }
}

fn main() {
    gpui_platform::application()
        .with_assets(Assets)
        .run(|cx| {
            gpui_component::init(cx);
            cx.spawn(async move |cx| {
                    let example = cx.new(|cx| Example::new(window, cx));
                    cx.new(|cx| Root::new(example, window, cx))
                })?;
                Ok::<_, anyhow::Error>(())
            })
            .detach();
        });
}
```

**Before**

スクリーンショット 2026-07-27 1 34 44

**After**

スクリーンショット 2026-07-27 1 39 57

## Checklist

### DIFF
--- a/crates/ui/src/input/element.rs
+++ b/crates/ui/src/input/element.rs
@@ -1637,39 +1637,71 @@ impl Element for TextElement {
             strikethrough: None,
         };
 
-        let runs = if let Some(highlight_styles) = highlight_styles.filter(|_| !is_empty) {
+        let runs = if let (false, Some(highlight_styles)) = (is_empty, highlight_styles) {
             let mut runs = Vec::with_capacity(highlight_styles.len() + 2);
 
             for (range, style) in &highlight_styles {
-                let mut highlighted_run = text_style.clone().highlight(*style).to_run(range.len());
+                let mut run = text_style.clone().highlight(*style).to_run(range.len());
                 if disabled {
-                    highlighted_run.color = highlighted_run.color.opacity(0.5);
+                    run.color = run.color.opacity(0.5);
                 }
-                extend_runs_with_ime_underline(
-                    &mut runs,
-                    highlighted_run,
-                    range.clone(),
-                    state
-                        .ime_marked_range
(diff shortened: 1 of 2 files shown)


huacnlee:
I pushed a follow-up commit to keep this PR focused on its original goal: fixing the IME composition underline.

The changes to `InputState::replace_and_mark_text_in_range` were removed because they alter core UTF-16 selection/caret behavior and are not required for the underline rendering fix. Even if that existing behavior needs correction, it should be handled separately with a clear rationale and focused tests.

I also replaced the mutable `push_run` closure with a small named helper and added regression coverage for both plain text and composition ranges that partially overlap or cross syntax-highlight ranges.

For future updates, please make changes only as needed for the stated goal and keep the diff as small and localized as possible. Avoid modifying unrelated core behavior in the same PR. If you discover a separate issue while working on a fix, please raise it independently rather than including it opportunistically.

PR #2613, opened 2026-07-29
input: Fix IME selection range during composition

zzfn:
## Summary

- interpret `new_selected_range_utf16` relative to the newly inserted marked text
- anchor both selection endpoints at the replacement start
- add a regression test for consecutive pinyin composition after a non-ASCII prefix

## Problem

During an IME composition update, the selected range is relative to the inserted marked text. The previous calculation converted it against the whole document and added the replacement end to the selection end. On the second pinyin character, this could produce an invalid or non-collapsed selection and move the IME candidate window to the beginning of the input.

## Testing

- `cargo test -p gpui-component --lib` (340 passed)
- `cargo clippy -p gpui-component --lib --tests -- --deny warnings`
- `rustfmt --edition 2024 --check crates/ui/src/input/state.rs`
- `git diff --check`

## AI assistance

Codex assisted with identifying the range conversion issue, implementing the fix, and drafting the regression test and PR description. The resulting diff and test behavior were reviewed and verified locally.

### DIFF
--- a/crates/ui/src/input/state.rs
+++ b/crates/ui/src/input/state.rs
@@ -3048,8 +3048,11 @@ impl EntityInputHandler for InputState {
             self.ime_marked_range = Some((range.start..range.start + new_text.len()).into());
             self.selected_range = new_selected_range_utf16
                 .as_ref()
-                .map(|range_utf16| self.range_from_utf16(range_utf16))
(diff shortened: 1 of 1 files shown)


madcodelife:
Thank you.

PR #2619, opened 2026-07-29
editor: add clickable breakpoint gutter

gintsgints:
## Summary

Adds an optional clickable breakpoint gutter to the multi-line editor (`InputState`), rendered as dots to the left of the line numbers — like the debug gutter in a code editor.

- Set breakpoints render as a filled dot (theme `danger`).
- Hovering the gutter reveals faint dots on every line so any line can be clicked to add a breakpoint.
- Clicking a dot toggles the breakpoint and emits `InputEvent::BreakpointToggled(line)` (0-based buffer row).

## API

`InputState`:
- `breakpoints_enabled(bool)` — builder to enable the gutter (requires line numbers).
- `set_breakpoints_enabled(bool, cx)` / `is_breakpoints_enabled()` — runtime toggle + getter.
- `breakpoints()` / `set_breakpoints(set, cx)` — read/replace the breakpoint lines.
- `toggle_breakpoint(line, cx)` — toggle a single line.

`InputEvent::BreakpointToggled(usize)` — emitted on toggle.

## Notes

Gutter width reserves `BREAKPOINT_DOT_WIDTH` only when enabled, so existing editors are unaffected. Layout/paint mirrors the existing fold-icon path (`layout_fold_icons`).

AI-assisted; reviewed and refactored to match project style.

### DIFF
--- a/crates/story/src/stories/input_story.rs
+++ b/crates/story/src/stories/input_story.rs
@@ -418,6 +418,7 @@ impl InputStory {
             }
             InputEvent::Focus => println!("Focus"),
             InputEvent::Blur => println!("Blur"),
+            InputEvent::BreakpointToggled(line) => println!("BreakpointToggled: {}", line),
         };
     }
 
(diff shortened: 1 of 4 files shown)


huacnlee (changes requested):
I don't think this is a good API design.

gintsgints:
> I don't think this is a good API design.

You think that gutter idea is bad at all or you think realisation need improvement?

PR #2627, opened 2026-07-31
select: Reset search before setting selected value

### LINKED ISSUE #2522: Reset search filter when set_selected_value is used on Select.
I had an issue where I wanted to set value of the select programmatically  and value is not visible/available because the previous search query/filter still applied on the set value.
There might be a different approach o a new public function it doesn't necessarily need to be reset on set value called.

Thanks.

lurenjia534:
Closes #2522

## Description

`SelectState::set_selected_value` previously looked up the requested value against the delegate's current, potentially filtered items. When an active search hid that value, `position` returned `None`, so the programmatic selection was cleared and the trigger did not display the requested item.

This change:

- makes `ListState::set_query` run the search delegate, matching its existing documented contract;
- centralizes user-entered and programmatic searches in the same task lifecycle;
- cancels a superseded search before starting its replacement;
- resets an active Select search and waits for the delegate to restore its items before resolving the requested value;
- preserves the existing synchronous `set_selected_value` behavior when no search is active;
- preserves the existing cursor behavior for ordinary List searches; and
- ensures the latest programmatic selection wins when an asynchronous reset is replaced.

The completion hook used by Select is crate-private. This does not add a new public API or emit a user selection event from the programmatic setter.

## Screenshot

Both screenshots use the same `SelectState` and unchanged item source. The Select is first filtered to `Rust`; `set_selected_value("Go")` is then called programmatically.

| Before | After |
| ------ | ----- |
| Before: the active Rust filter prevents the Go value from being selected | After: the filter is reset and the Go value is selected and rendered |

Before the fix, the `Rust` filter remains active, the committed value is empty, and the trigger still shows its placeholder. After the fix, the query is cleared, all rows are restored, `Go` is committed, and the trigger renders `Go`.

## How to Test

[GitHub Actions](https://github.com/longbridge/gpui-component/actions/runs/30601643916) passed on Linux, Windows, and macOS. The new regression coverage is not ignored or platform-gated.

The new regression cases were first run against upstream commit `88f102d13654fe25aa2fede076274b6b751a3704` without the fix. Five relevant paths failed as expected: synchronous filtered selection, grouped selection, asynchronous/latest-call selection, real Select interaction, and programmatic `ListState::set_query`.

After the fix:

- `cargo test --locked --offline -p gpui-component select::tests` — 8 passed
- `cargo test --locked --offline --all` — passed, including all 366 `gpui-component` tests and workspace/example/doc tests
- `cargo clippy --locked --offline --workspace --all-targets -- --deny warnings` — passed
- `cargo check --locked --offline -p gpui-component --no-default-features` — passed
- `cargo check --locked --offline -p gpui-component --no-default-features --target wasm32-unknown-unknown` — passed
- `rustfmt --check --config-path .rustfmt.toml crates/ui/src/list/list.rs crates/ui/src/select.rs` — passed
- Manual Linux story reproduction — passed; before/after screenshots captured

The automated coverage includes:

- the original filtered Select regression;
- public programmatic List search;
- unchanged synchronous behavior without a filter;
- grouped index restoration;
- an asynchronous delegate with cancellation and latest-call-wins behavior; and
- an end-to-end `VisualTestContext` interaction that dispatches the actual trigger click, edits the actual query `Input`, applies a programmatic value, and verifies both state and rendered trigger output.

The first test-only adjustment (`9f61e43`) still failed on macOS because `VisualTestContext::simulate_click` runs the executor after mouse-up. That allowed the focused query `Input` to redraw and request `TestWindow::window_handle` before focus could be restored.

The final test change (`4170a4a`) dispatches mouse-down/up in one update, restores trigger focus before executor refresh, and edits the actual query `Input` through its normal `InputEvent::Change` path. No production code was changed for this synthetic `TestWindow` limitation, no platform test was skipped, and the latest Linux, Windows, and macOS CI run passed.

## AI Assistance

AI assisted with implementation and test scaffolding. The contributor reviewed the final diff, generated code, and test evidence before marking this PR ready.

## Checklist

- [x] Passed the related story reproduction on Linux.
- [x] Not platform-specific; no platform-specific rendering or performance path was changed.

### DIFF
--- a/crates/ui/src/list/list.rs
+++ b/crates/ui/src/list/list.rs
@@ -73,6 +73,7 @@ pub struct ListState<D: ListDelegate> {
     options: ListOptions,
     delegate: D,
     last_query: Option<String>,
+    search_pending: bool,
     scroll_handle: VirtualListScrollHandle,
     rows_cache: RowsCache,
     selected_index: Option<IndexPath>,
@@ -105,6 +106,7 @@ where
             rows_cache: RowsCache::default(),
             query_input,
             last_query: None,
+            search_pending: false,
             selected_index: None,
             selectable: true,
             searchable: false,
@@ -214,6 +216,27 @@ where
     /// Set the query text of the search input, this will trigger a search.
     pub fn set_query(&mut self, query: &str, window: &mut Window, cx: &mut Context<Self>) {
         let query = query.to_string();
+        self.set_query_input(query.clone(), window, cx);
+        self.start_search(query, None, window, cx);
+    }
+
+    pub(crate) fn is_search_pending(&self) -> bool {
+        self.search_pending
+    }
+
+    pub(crate) fn set_query_with_completion(
+        &mut self,
+        query: &str,
+        on_complete: impl FnOnce(&mut Self, &mut Window, &mut Context<Self>) + 'static,
+        window: &mut Window,
+        cx: &mut Context<Self>,
+    ) {
+        let query = query.to_string();
+        self.set_query_input(query.clone(), window, cx);
+        self.start_search(query, Some(Box::new(on_complete)), window, cx);
+    }
+
+    fn set_query_input(&mut self, query: String, window: &mut Window, cx: &mut Context<Self>) {
         self.query_input.update(cx, |input, cx| {
             input.set_value(query, window, cx);
         });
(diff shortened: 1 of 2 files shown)


huacnlee:
Close by https://github.com/longbridge/gpui-component/pull/2670 fixed.

You have change too many other codes.

lurenjia534:
> Close by #2670 fixed.
> 
> You have change too many other codes.

Thank you. I made it too complicated (over-engineered it). The changes should have been smaller and more focused. Thank you for the review.

PR #2156, opened 2026-03-13
text_view: Fix clipped markdown hit testing

suxiaoshao:
Closes #2155

## Description

Fix clipped-region hit testing for markdown content rendered by `TextView`. `TextView` now uses a hitbox-aware check when starting or clearing selection, and `Inline` link clicks use the same clipped-region-aware hit testing instead of raw bounds checks. This prevents hidden markdown from remaining interactive when it is clipped by `overflow_hidden()` or covered by sibling content.

### Before 

https://github.com/user-attachments/assets/77d5a9f1-53c3-47c3-a958-bc1806630dcd

### After 

https://github.com/user-attachments/assets/67144c0f-08ad-4c90-9636-9a68ab8934bd

## How to Test

Run:

```bash
cargo test -p gpui-component clipped_markdown_ -- --nocapture
```

Then verify that clicking or dragging over a visually clipped markdown region no longer starts selection or opens a link.

## Checklist

- [ ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) document and followed the guidelines.
- [x] Reviewed the changes in this PR and confirmed AI generated code (If any) is accurate.
- [ ] Passed `cargo run` for story tests related to the changes.
- [ ] Tested macOS, Windows and Linux platforms performance (if the change is platform-specific)

PR #2628, opened 2026-07-31
markdown: Support copying the Markdown source of a selection

### LINKED ISSUE #1244: text_view: Support selection to output markdown.
For AI

stippi:
Closes #1244

## Description

`TextView` now has a `SelectionFormat`, so a selection can be copied as its Markdown source instead of the rendered text.

`SelectionFormat::Plain` is the default and keeps the current behavior. With `SelectionFormat::Source`, `TextViewState::selected_text` (what copy writes to the clipboard) reconstructs the Markdown of the selection: inline marks are re-wrapped (`**bold**`, `` `code` ``, `[text](url)` …), headings keep their `#` prefix, list items keep their marker, code blocks keep their fence, and top-level blocks are separated by a blank line. Select-all returns the original source verbatim.

```rust
gpui_component::markdown(source)
    .selectable(true)
    .selection_format(SelectionFormat::Source)
```

## Screenshot

I don't have meaningful before and after, but in the screenshot below, I intentionally selected within the heading and within some mark, used `cmd-c` and pasted into MacDown:

image

## How to Test

Just select text and copy as before, but it works only on views with `SelectionFormat::Source`.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -693,7 +693,7 @@ dependencies = [
  "bitflags 2.11.1",
  "cexpr",
  "clang-sys",
- "itertools 0.11.0",
+ "itertools 0.13.0",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -1197,7 +1197,7 @@ dependencies = [
 [[package]]
 name = "collections"
 version = "0.1.0"
-source = "git+https://github.com/zed-industries/zed#4aad57fd1f002f9feeea2b7fb6229ccbcd576cb1"
+source = "git+https://github.com/zed-industries/zed#cc053a4a6fa2fd0e8793201ed9099466af1be0b1"
 dependencies = [
  "gpui_util",
  "indexmap",
@@ -1644,9 +1644,9 @@ dependencies = [
 
 [[package]]
 name = "ctor"
-version = "1.0.7"
+version = "1.0.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "01334b89b69ff726750c5ce5073fc8bd860e99aa9a8fc5ca11b04730e3aee97a"
+checksum = "2d83cb7e7a873830708d6b02a78cd36a592c6fa14bf267b68725103b85c0d77f"
 dependencies = [
  "link-section",
  "linktime-proc-macro",
@@ -1753,7 +1753,7 @@ dependencies = [
 [[package]]
 name = "derive_refineable"
 version = "0.1.0"
-source = "git+https://github.com/zed-industries/zed#4aad57fd1f002f9feeea2b7fb6229ccbcd576cb1"
+source = "git+https://github.com/zed-industries/zed#cc053a4a6fa2fd0e8793201ed9099466af1be0b1"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2874,7 +2874,7 @@ dependencies = [
  "log",
  "presser",
  "thiserror 2.0.18",
- "windows 0.58.0",
+ "windows 0.62.2",
 ]
 
 [[package]]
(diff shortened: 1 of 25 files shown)


huacnlee (changes requested):
Thanks for this! I pushed a few changes on top:

- The opt-in is now a SelectionFormat on TextView (Plain by default, Source for the Markdown source), so selected_text() stays the one path — no parallel selected_source() chain through Root, WindowExt and TextViewState.
- The markdown and html examples have a status bar toggle for it.
- Fixed inline images: a paragraph stores one InlineState per run of children between images, so mapping a selection offset against every child made the text before an image come out again as the text after it.

Testing it further, though, the source copy only covers part of a document. Paragraphs, headings, flat lists and code blocks work, but these do not:

- Images — a paragraph that is only an image copies as nothing.
- Tables — cells are joined with spaces instead of being piped back into a Markdown table, and the alignment row is missing.
- Nested lists — flattened to one level, the indent is lost.

Blockquotes lose their > prefix and task list items lose their checkbox as well.

The feature promises the Markdown source of a selection, so it should round-trip the whole document rather than a subset. Could you finish the remaining node types and add a test per type? Once each one round-trips I'm happy to merge.

stippi:
> Thanks for this! I pushed a few changes on top:
> 
> * The opt-in is now a SelectionFormat on TextView (Plain by default, Source for the Markdown source), so selected_text() stays the one path — no parallel selected_source() chain through Root, WindowExt and TextViewState.
> * The markdown and html examples have a status bar toggle for it.
> * Fixed inline images: a paragraph stores one InlineState per run of children between images, so mapping a selection offset against every child made the text before an image come out again as the text after it.
> 
> Testing it further, though, the source copy only covers part of a document. Paragraphs, headings, flat lists and code blocks work, but these do not:
> 
> * Images — a paragraph that is only an image copies as nothing.
> * Tables — cells are joined with spaces instead of being piped back into a Markdown table, and the alignment row is missing.
> * Nested lists — flattened to one level, the indent is lost.
> 
> Blockquotes lose their > prefix and task list items lose their checkbox as well.
> 
> The feature promises the Markdown source of a selection, so it should round-trip the whole document rather than a subset. Could you finish the remaining node types and add a test per type? Once each one round-trips I'm happy to merge.

Yes, happy to follow up on it. In code-assistant, I also noticed that the copied text is sometimes missing large chunks when I select across multiple views that each display their chunk of the markdown. Might be due to the virtualized list in code-assistant specifically, but maybe there is a bug somewhere still. Thanks for taking a look and also fixing the double code-path thing. That annoyed me, too.

stippi:
The remaining nodes are done. Claude said a single image at the selection start or end could not participate in the selection, since there are additional changes needed to make image views selectable. It suggested doing that in a follow up PR. What's your take?

stippi:
I've reviewed the code again, it looks clean to me. There is a new test in text_view.rs which just guards against a regression for being able to select clipped text. Let me know if you want to (not) keep that.
I let Claude investigate selecting "just image" paragraphs at the start or end of a selection. If that turns out less than expected, it could still be part of this PR.

stippi:
Claude's analysis for the image gap (happily copied straight from code-assistant using this PR's feature):

## What it would take (short version)

The gap has four parts, all because images aren't first-class selection participants:

1. A selection can't **start or end** on an image (endpoint resolution only anchors to text `Inline` bounds via `inside_text`), so a drag over *just* an image sets `did_hit_text = false` and is discarded.
2. Image blocks never report `has_selection()`, so they can only be picked up by the "enclosed between two selected blocks" heuristic — a **leading or trailing** image in a selection is dropped.
3. Nothing registers image geometry, so the window-selection band can't test whether it covers an image.
4. **Plain** mode emits nothing for images even when enclosed (only Source has the image logic).

Closing it means making an image-only paragraph a geometric selection participant: register its painted bounds (like `register_selectable_text_inline`), let an endpoint anchor to it (`inside_text`/a new `inside_image`), mark the block "selected" when the band covers its rectangle during paint, and emit `alt` (Source) / alt-or-nothing (Plain) from `ParsedDocument::selected_text` for covered image blocks — while preserving the existing enclosure behavior and not regressing #2156/#2664.

---

I would make it a follow up PR.

huacnlee:
1. I have fixed in main branch.
2. I remembered that I also fixed too.

huacnlee (approved):

PR #2629, opened 2026-07-31
input: Support configurable wrapping indent

hlcfan:
## Description

- Support `WrappingIndent`
  - Some: Wrapped lines get the same indentation as the parent
  - None: No indentation. Wrapped lines begin at column 1
- Fix the current long text rendering with `soft_wrap` enabled
  - Add left offset on continuation wrapped lines
  - Fix selection corner segments

## Screenshot

**Before**
Before

**After**
| Same                       | None                       |
| Screenshot 2026-08-04 at 10 52 15 PM | After |

## Break Changes

Set the default wrapping indent to `Same`.

## How to Test

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce.

## Checklist

### DIFF
--- a/crates/ui/src/input/display_map/text_wrapper.rs
+++ b/crates/ui/src/input/display_map/text_wrapper.rs
@@ -10,6 +10,16 @@ use sum_tree::{Bias, Dimensions, SumTree};
 
 use crate::input::{LastLayout, Point as TreeSitterPoint, RopeExt, WhitespaceIndicators};
 
+/// Controls how soft-wrapped continuation lines are indented.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum WrappingIndent {
+    /// Continuation lines start flush-left at the full editor width.
+    #[default]
+    None,
+    /// Continuation lines keep the same indentation as the first line.
+    Same,
+}
+
 /// A line with soft wrapped lines info.
 #[derive(Debug, Clone)]
 pub(crate) struct LineItem {
@@ -126,6 +136,7 @@ pub(crate) struct TextWrapper {
     font_size: Pixels,
     /// If is none, it means the text is not wrapped
     wrap_width: Option<Pixels>,
+    wrapping_indent: WrappingIndent,
     /// The lines by split \n
     pub(crate) lines: SumTree<LineItem>,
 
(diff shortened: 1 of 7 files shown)


crates/ui/src/input/state.rs
```diff
@@ -963,6 +967,14 @@ impl InputState {
         self
     }
 
+    /// Set how soft-wrapped continuation lines are indented, default is [`WrappingIndent::None`]
+    /// (flush-left at the fiull editor width).
```
madcodelife: `fiull` typo?

crates/ui/src/input/display_map/text_wrapper.rs
```diff
@@ -10,6 +10,16 @@ use sum_tree::{Bias, Dimensions, SumTree};
 
 use crate::input::{LastLayout, Point as TreeSitterPoint, RopeExt, WhitespaceIndicators};
 
+/// Controls how soft-wrapped continuation lines are indented.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum WrappingIndent {
+    /// Continuation lines start flush-left at the full editor width.
+    #[default]
+    None,
+    /// Continuation lines keep the same indentation as the first line.
```
madcodelife, line 18: The doc says "continuation lines keep the same indentation as the first line", but in `LineLayout::paint` (text_wrapper.rs:656) every sub-line is painted at x = 0:

    line.paint(pos + point(px(0.), ix * line_height), ...)

Only y advances. What `Same` really does is make gpui reserve the indent width while wrapping, so continuation lines break early and then render flush-left. Should the doc comment be updated to match?

hlcfan: Thanks for pointing this out, I was thinking to check this behavior in another PR. Now the rendering is fixed for the `Same` mode, but the text selection isn't drawing correctly. Will check this tomorrow.

image

hlcfan: This is fixed.

hlcfan:
@madcodelife can you help take another look? Thanks.

madcodelife:
Sorry for the delay, looks great now. Thank you!

PR #2637, opened 2026-08-02
button: Only expose toggled state on toggle buttons

### LINKED ISSUE #2635: Button exposes selection state instead of toggle state
### Description

`Button` always writes `aria_selected(false)`, including for ordinary push buttons. When a caller uses `.selected(...)`, the component still exposes selection state even though the public API and rendered control behave as a toggle button.

Accessibility clients therefore receive the wrong state model: ordinary buttons advertise a selection value, while configured toggle buttons do not advertise AccessKit `Toggled` state.

### Environment

- GPUI: git revision `1a246efd`
- GPUI Component: `0.5.2`, main `88f102d1`
- Platform: cross-platform AccessKit metadata; reproduced in the rendered-element test harness

### Steps to Reproduce

1. Render an ordinary `Button` without calling `.selected(...)`.
2. Inspect its emitted AccessKit node.
3. Render buttons with `.selected(false)` and `.selected(true)`.
4. Inspect their emitted AccessKit nodes.

### Screenshots

Not applicable. Rendering does not change.

### Expected

An ordinary button emits no toggle state. Calling `.selected(false)` emits `Toggled::False`; calling `.selected(true)` emits `Toggled::True`.

### Actual

Every button emits selection state, and `.selected(...)` does not emit the matching toggle state.

railapex:
Closes #2635

### Description

Track whether the caller explicitly configured `.selected(...)`. Emit AccessKit `Toggled::False` or `Toggled::True` only in that case. Ordinary buttons retain their button role and press action without advertising toggle state.

This changes only accessibility metadata. It does not change selected styling or the public `.selected(...)` API.

### Screenshot

Not applicable. Rendering does not change.

### How to Test

Exact candidate `e4720a3b` on Rust 1.97.1:

- `cargo test -p gpui-component button --lib`: 9 passed
- `cargo test -p gpui-component`: 344 passed
- `cargo clippy -p gpui-component --all-targets -- -D warnings`
- story application compile: passed

The rendered-element regression writes the actual AccessKit node and proves ordinary `None`, explicit false `Toggled::False`, and explicit true `Toggled::True`.

### AI Assistance

The implementation and tests were AI-assisted by Friday. Chris Hynes defined the behavior and owns the final review.

### Checklist

- [ ] Passed `cargo run` for story tests related to the changes. The story application compiles; it has not been launched.
- [ ] Tested macOS, Windows and Linux platforms performance (if the change is platform-specific). This is a cross-platform metadata change with rendered-node unit coverage, not live platform E2E coverage.

### DIFF
--- a/crates/ui/src/button/button.rs
+++ b/crates/ui/src/button/button.rs
@@ -1211,6 +1211,78 @@ mod tests {
     }
 
     #[gpui::test]
+    fn test_button_emits_toggle_state_only_when_selected_is_configured(
+        cx: &mut gpui::TestAppContext,
+    ) {
+        use crate::ElementExt as _;
+        use gpui::{Element as _, IntoElement as _, Render};
+        use std::sync::{Arc, Mutex};
+
+        struct ButtonA11yProbe {
+            states: Arc<Mutex<Vec<Option<gpui::accesskit::Toggled>>>>,
+        }
+
+        impl Render for ButtonA11yProbe {
+            fn render(
+                &mut self,
+                _window: &mut Window,
+                _cx: &mut gpui::Context<Self>,
+            ) -> impl IntoElement {
+                let states = self.states.clone();
+                div().on_prepaint(move |_, window, cx| {
+                    let ordinary = Button::new("ordinary")
+                        .label("Ordinary")
+                        .render(window, cx)
+                        .into_element();
+                    let mut ordinary_node = gpui::accesskit::Node::new(Role::Button);
+                    ordinary.write_a11y_info(&mut ordinary_node);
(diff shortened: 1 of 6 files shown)


madcodelife:
Thanks for the PR!

`.selected()` is really the styling switch here, and wrappers like `Popover::trigger` call it unconditionally, so plain popover triggers were picking up toggle state too.

So I've split them apart. selected is styling only now, and there's a new Button::toggled() for the a11y side.

railapex:
Awesome, thanks! Finding GPUI very quick and useful.

This and a couple of my other PR's add a couple little touches that will close a couple of a11y edge case gaps and thus help both humans and agent test drivers.

PR #2643, opened 2026-08-03
input: Add aria_label to set the accessible name

mike-marcacci:
## Description

`Input` emitted a role-bearing AccessKit node with no way to name it, so assistive technology announced every input as an unnamed text field.

## How to Test

Use a screen reader to announce an `Input` element.

## Checklist

- [ ] ~Passed `cargo run` for story tests related to the changes.~ **Not needed.**
- [ ] ~Tested macOS, Windows and Linux platforms performance (if the change is platform-specific)~ **Only tested on macOS.**

### DIFF
--- a/crates/ui/src/input/input.rs
+++ b/crates/ui/src/input/input.rs
@@ -429,11 +429,24 @@ impl RenderOnce for Input {
             && state.mode.is_single_line();
         let has_suffix = suffix.is_some() || state.loading || self.mask_toggle || show_clear_button;
 
+        let placeholder = Some(state.placeholder.clone()).filter(|p| !p.is_empty());
+
+        // Don't use a mask-derived placeholder ("(___)___-___") as an aria_label fallback.
+        let placeholder_is_mask =
+            state.mask_pattern.placeholder().as_deref() == placeholder.as_deref();
+
+        let aria_label = match self.aria_label {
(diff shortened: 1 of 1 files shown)


crates/ui/src/input/input.rs
```diff
@@ -425,6 +432,9 @@ impl RenderOnce for Input {
         div()
             .id(("input", self.state.entity_id()))
             .role(accessibility_role)
+            .when_some(self.aria_label.clone(), |this, label| {
```
madcodelife: The clone is unnecessary.

madcodelife:
Thanks for the PR!

Two things:
1. Can we default `aria_label` to the placeholder when it's not set? Every call site has to opt in otherwise, so most inputs would still have no label. Like [zed](https://github.com/zed-industries/zed/blob/2268045a119030735f762e5afaf59da0bda869f4/crates/settings_ui/src/components/input_field.rs#L220).
2. Would be good to set `aria_placeholder` too.

mike-marcacci:
Great calls. Rebased, fixed, and added placeholder handling. I avoided falling back to the placeholders that are mask-derived, since a label like `(___)___-___` probably wouldn't be very useful.

madcodelife:
Thanks.

PR #2644, opened 2026-08-03
button, checkbox: Add RoleOverride; and ability to specify Presentational

mike-marcacci:
## Description

This extends `Input`'s existing `role` override to `Button` and `Checkbox`, and lets all three take `RoleOverride::Presentational` — the HTML `role="presentation"` or `role="none"` analog where the component instance emits no `AccessKit`/etc node of its own.

This solves uses that annotate a wrapping element as the accessible control (a labelled row with an embedded remove button, a labelled field wrapping an input) which otherwise present two nodes per control to assistive technology.

## How to Test

Override a role and navigate with a screen reader or similar assistive technology.

## Break Changes

This should not break any existing uses, as `role: impl Into<Option<Role>>` will still accept any concrete `Role` passed to an existing `Input`'s `.role(r)`. It's unlikely but theoretically possible that `.role(x.into())` has been used by some consuming app, which would become ambiguous.

## Checklist

- [ ] ~Passed `cargo run` for story tests related to the changes.~ **Not needed.**
- [ ] ~Tested macOS, Windows and Linux platforms performance (if the change is platform-specific)~ **Only tested on macOS.**

### DIFF
--- a/crates/ui/src/button/button.rs
+++ b/crates/ui/src/button/button.rs
@@ -1,8 +1,8 @@
 use std::rc::Rc;
 
 use crate::{
-    ActiveTheme, Colorize as _, Disableable, FocusableExt as _, Icon, Selectable, Sizable, Size,
-    StyleSized, StyledExt,
+    ActiveTheme, Colorize as _, Disableable, FocusableExt as _, Icon, RoleOverride, Selectable,
+    Sizable, Size, StyleSized, StyledExt,
     button::ButtonIcon,
     h_flex,
     select::Caret,
@@ -192,6 +192,7 @@ pub struct Button {
     disabled: bool,
     pub(crate) selected: bool,
     toggled: Option<bool>,
+    role: RoleOverride,
     variant: ButtonVariant,
     rounded: ButtonRounded,
     outline: bool,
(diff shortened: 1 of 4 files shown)


madcodelife:
Thanks! The role override itself is useful.

One thing on the API shape though: could we use a three-state enum instead of `Option<Option<Role>>`?

```rust 
#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
pub enum RoleOverride {
    /// Use the component's default role.
    #[default]
    Inherit,
    /// Emit no accessibility node, so an ancestor can carry the role.
    Presentational,
    /// Use this role instead of the default.
    Role(Role),
}

impl From<Role> for RoleOverride {
    fn from(role: Role) -> Self {
        Self::Role(role)
    }
}

impl From<Option<Role>> for RoleOverride {
    fn from(role: Option<Role>) -> Self {
        match role {
            Some(role) => Self::Role(role),
            None => Self::Presentational,
        }
    }
}
```

mike-marcacci:
Great call. Rebased and refactored with a `RoleOverride` enum.

I chose to use `Implicit` instead of `Inherit` because the latter has a very different meaning in the CSS/HTML world, and I didn't want to invite confusion.

madcodelife:
#2643 After merging, there are conflicts that need to be resolved.

mike-marcacci:
Hi @madcodelife, my apologies for the delay! I fixed those conflicts, so this should be good to go 👍🏼

PR #2663, opened 2026-08-08
combobox: expose search query accessors

ItsCacia:
## Summary

- add `ComboboxState::query()` to read the current search query
- add `ComboboxState::set_query()` to update the query and immediately refresh filtered items
- keep programmatic queries consistent with typed input by trimming only for filtering while preserving the displayed text
- update the Combobox footer example to read the live query when adding a new university
- add regression coverage for both the query text and filtered results

## Testing

- `cargo test -p gpui-component test_combo_box_set_query_updates_text_and_filters_items --lib`
- `cargo test -p gpui-component combobox --lib`
- `cargo test -p gpui-component --lib`
- `cargo check -p gpui-component-story`

## Manual verification

1. Run the Combobox story.
2. Open the **Footer** example.
3. Type a university name.
4. Click **New university** and confirm the typed query is added to the options.


crates/ui/src/combobox.rs
```diff
@@ -370,6 +370,25 @@ where
...
+            list.query_input.update(cx, |input, cx| {
+                input.set_value(value, window, cx);
+            })
+        });
+    }
+
+    /// Clears the search query.
+    pub fn clear_search_value(&self, window: &mut Window, cx: &mut App) {
+        self.set_search_value("", window, cx);
+    }
```
huacnlee: Remove this method.

ItsCacia: Removed.

PR #2700, opened 2026-08-13
input: Clip selection ranges to UTF-8 boundaries

### LINKED ISSUE #2699: InputBaseState::set_selected_range accepts non-character byte offsets and can panic on copy
## Description

`InputBaseState::set_selected_range` accepts UTF-8 byte offsets but only clamps them to the text byte length. It does not ensure that the offsets are valid UTF-8 character boundaries.

For text containing a multi-byte character, a caller can therefore store an invalid selection range. A later operation such as copy or cut passes that range directly to `Rope::slice`, which panics when an endpoint is not on a character boundary.

This is especially relevant when selection offsets originate from external sources such as search indexes, LSP responses, persisted state, or other APIs using a different offset encoding.

The issue is reproducible on the latest `main` commit `dc2943fac0ee9b907e21cc415867bd45615c220c`.

## Environment

- GPUI: v0.2.2, revision `cc053a4a6fa2fd0e8793201ed9099466af1be0b1`
- GPUI Component: `main` at `dc2943fac0ee9b907e21cc415867bd45615c220c`
- Platform: Arch Linux, Linux 7.1.7, x86_64
- Display server: native Wayland session (`WAYLAND_DISPLAY=wayland-0`)
- Rust: rustc 1.96.1
- Reproduction type: native GPUI window and event loop, not a unit-test harness

## Steps to Reproduce

Add the following native example as `crates/base/examples/selection_byte_offset_gui_repro.rs`:

```rust
use std::time::Duration;

use gpui::{
    App, AppContext as _, Context, Entity, IntoElement, ParentElement as _,
    Render, Styled as _, Window, WindowBounds, WindowOptions, div, px, size,
};
use gpui_base::input::{Copy, Input, InputState};

struct SelectionCrashRepro {
    input: Entity<InputState>,
}

impl SelectionCrashRepro {
        let input = cx.new(|cx| InputState::new(window, cx).default_value("éx"));
        let repro_input = input.clone();

        window
            .spawn(cx, async move |cx| {
                // Let the native window render and register its action
                // handlers first.
                smol::Timer::after(Duration::from_secs(2)).await;

                repro_input
                    .update_in(cx, |state, window, cx| {
                        state.focus(window, cx);
                        let base = state.base_state().clone();
                        base.update(cx, |state, cx| {
                            // `é` occupies bytes 0..2, so 1 is not a
                            // UTF-8 character boundary.
                            state.set_selected_range(0..1, cx);
                            eprintln!(
                                "stored selection {:?} for text {:?}",
                                state.selected_range(),
                                state.value()
                            );
                        });

                        // Dispatch through the focused window's normal Copy
                        // action handler, as a Ctrl+C key binding would.
                        window.dispatch_action(Box::new(Copy), cx);
                    })
                    .expect("repro window closed before Copy dispatch");
            })
            .detach();

        Self { input }
    }
}

impl Render for SelectionCrashRepro {
    fn render(
        &mut self,
        _: &mut Window,
        _: &mut Context<Self>,
    ) -> impl IntoElement {
        div()
            .size_full()
            .p_8()
            .bg(gpui::rgb(0xffffff))
            .child(Input::new(&self.input))
    }
}

fn main() {
    gpui_platform::application().run(|cx: &mut App| {
        gpui_base::init(cx);

        let options = WindowOptions {
            window_bounds: Some(WindowBounds::centered(
                size(px(620.), px(220.)),
                cx,
            )),
            ..WindowOptions::default()
        };

        cx.open_window(options, |window, cx| {
            cx.new(|cx| SelectionCrashRepro::new(window, cx))
        })
        .expect("failed to open native GPUI repro window");
        cx.activate(true);
    });
}
```

Run:

```bash
RUST_BACKTRACE=1 cargo run -p gpui-base \
  --example selection_byte_offset_gui_repro
```

The native window opens and renders the input. After two seconds, the focused
Copy action is dispatched and the main GUI process exits with status `101`.

## Screenshots

Not applicable. The relevant output from the native Wayland run is:

```text
stored selection 0..1 for text "éx"

thread 'main'
panicked at crates/base/src/input/base/state.rs:2135:39:
byte index is not on a char boundary

stack backtrace:
ropey::rope::Rope::slice
gpui_base::input::state::InputBaseState::copy
gpui::window::Window::dispatch_action_on_node_inner
gpui_linux::linux::wayland::client::WaylandClient::run
gpui::app::Application::run

process exit status: 101
```

## Expected

The public selection setter should not allow an invalid UTF-8 byte range to be stored.

Offsets should either be normalized to character boundaries or rejected. Collapsed selections should remain collapsed after normalization. Subsequent operations such as copy and cut should not panic.

## Actual

`set_selected_range(0..1)` on the text `"éx"` stores `0..1` unchanged. Calling `copy` then passes this range to `Rope::slice`, which panics because byte offset `1` lies inside the two-byte UTF-8 encoding of `é`.

## Impact

The demonstrated impact is an availability failure in the embedding application. Exploitation requires a caller to pass a non-character-boundary offset to the public setter, but once such a range is stored, ordinary selection consumers such as copy or cut can deterministically panic the process.

lurenjia534:
Closes #2699

## Description

`InputBaseState::set_selected_range` previously clamped offsets only to the text byte length. An offset inside a multi-byte UTF-8 character could therefore be stored in `selected_range`. Copying or cutting that selection later passed the invalid range to `Rope::slice` and panicked the application.

This change clips non-empty selection ranges outward to valid character boundaries before storing them. Empty ranges remain collapsed and are clipped to the preceding character boundary.

The patch is intentionally limited to the public selection setter and a regression test for the reported crash path. Internal cursor movement and selection helpers are unchanged.

## Screenshot

Not applicable. The change fixes non-visual selection state validation and a process panic.

## How to Test

```bash
cargo fmt --all -- --check
cargo test -p gpui-base
```

The regression test uses the text `"éx"` and selection `0..1`, where byte offset `1` is inside `é`. It verifies that the range becomes `0..2` and that the existing Copy action completes without panicking. It also verifies that an empty `1..1` range remains empty after clipping.

The original native Wayland GUI reproduction was also rerun after the patch. The focused Copy action completed without a panic, and the normalized selection was reported as `0..2`.

## AI Assistance

AI assisted with the implementation, regression test, native reproduction, and PR draft. The contributor reviewed the generated changes and test evidence before submission.

## Checklist

- [X] Reviewed the changes in this PR and confirmed AI-generated code is accurate.
- [x] Passed `cargo test -p gpui-base` (264 unit tests and 1 integration test).
- [x] Passed a native `cargo run` GPUI reproduction for the affected Copy action on Linux Wayland.
- [x] The change is not platform-specific; no platform-specific behavior was modified.

### DIFF
--- a/crates/base/src/input/base/state.rs
+++ b/crates/base/src/input/base/state.rs
@@ -2244,10 +2244,17 @@ impl InputBaseState {
     }
 
     /// Set the selected range using UTF-8 byte offsets.
+    ///
+    /// Non-empty ranges expand to character boundaries. Empty ranges remain empty and are
+    /// clipped to the preceding character boundary.
     pub fn set_selected_range(&mut self, range: Range<usize>, cx: &mut Context<Self>) {
-        let len = self.text.len();
-        let start = range.start.min(len);
-        let end = range.end.min(len);
+        let end_bias = if range.start == range.end {
+            Bias::Left
+        } else {
+            Bias::Right
+        };
+        let start = self.text.clip_offset(range.start, Bias::Left);
+        let end = self.text.clip_offset(range.end, end_bias);
 
         self.move_to(start, None, cx);
         self.selection_reversed = false;
@@ -3851,6 +3858,24 @@ mod tests {
                 // clamped + collapsed
                 s.set_selected_range(100..100, cx);
                 assert_eq!(s.selected_range(), 11..11);
+            });
+        });
+    }
+
+    #[gpui::test]
+    fn test_set_selected_range_clips_to_utf8_boundaries(cx: &mut TestAppContext) {
+        let input_view = InputView::build(cx, |state| state.default_value("éx"));
+        let mut cx = VisualTestContext::from_window(input_view.window_handle.into(), cx);
+        let input = input_view.input;
+
+        cx.update(|window, cx| {
+            input.update(cx, |state, cx| {
+                state.set_selected_range(0..1, cx);
+                assert_eq!(state.selected_range(), 0..2);
+                state.copy(&Copy, window, cx);
+
+                state.set_selected_range(1..1, cx);
+                assert_eq!(state.selected_range(), 0..0);
             });
         });
     }


huacnlee:
Thanks

PR #2715, opened 2026-08-14
input: Hide the editing engine and move LSP onto EditorState

huacnlee:
## Description

Stacked on #2714 — review that one first.

`InputBaseState` is the shared editing engine that `InputState`, `TextareaState`
and `EditorState` are built on. It was re-exported from `gpui_component::input`,
and because `EditorState` did not forward LSP or highlighting, reaching those
features meant dropping down to the engine. Four examples and the inspector did
exactly that.

This keeps the engine inside the framework and gives every control the state
that matches it. Everything the examples needed moves onto `EditorState`, where
it belongs.

`InputState`, `TextareaState` and `EditorState` keep every method they had, with
the same signatures — `value()`, `cursor_position(cx)`, `base_state()`,
`diagnostics_mut()`, `Input::from_base` all unchanged. The additions are purely
additive.

> AI-generated with Claude Code, reviewed and adjusted by hand.

### Control → state

| Control | State | Was |
|---|---|---|
| `inspector` rust/json editors | `EditorState` | `InputBaseState` |
| `examples/html`, `examples/markdown` | `EditorState` | `InputBaseState` |
| `examples/editor` main editor | `EditorState` | `InputBaseState` |
| `examples/large-text` body | `TextareaState` | `InputBaseState` |
| go-to-line fields | `InputState` | `InputBaseState` |

`large-text` was built with `multi_line(true)` rather than as a code editor, so
it is a textarea; `TextareaState` gains the `tab_size`, `set_cursor_position`,
`set_soft_wrap` and `set_placeholder` it needed for that.

### New readers borrow, they do not mirror

`text`, `cursor` and `cursor_position` read through to the engine and take a
`cx`. A cached copy would cost a second copy of the whole text and can go stale
— the cursor especially, since it moves without a text change and `InputEvent`
has no event for it. The existing `value()` cache is left exactly as it is.

## Breaking Changes

- `InputBaseState` is no longer exported from `gpui_component::input`. Use the
  state of the control you are building. It stays reachable through
  `base_state()` for code that still needs the engine, so a field only needs its
  type name updated.

```diff
- use gpui_component::input::{Input, InputBaseState};
- let state = cx.new(|cx| InputBaseState::new(window, cx).code_editor("rust"));
- Input::from_base(&state)
+ use gpui_component::input::{Editor, EditorState};
+ let state = cx.new(|cx| EditorState::new("rust", window, cx));
+ Editor::new(&state)
```

```diff
- input_state: Entity<InputBaseState>,
+ input_state: Entity<EditorState>,
```

  LSP providers are configured through `update_lsp`, so the field names stay the
  same:

```diff
- let input_state = editor_state.read(cx).base_state().clone();
- input_state.update(cx, |state, _| {
-     state.lsp.completion_provider = Some(lsp.clone());
-     state.lsp.hover_provider = Some(lsp.clone());
-     state.lsp.definition_provider = Some(lsp.clone());
- });
+ editor_state.update(cx, |state, cx| {
+     state.update_lsp(cx, |lsp| {
+         lsp.completion_provider = Some(lsp_store.clone());
+         lsp.hover_provider = Some(lsp_store.clone());
+         lsp.definition_provider = Some(lsp_store.clone());
+     });
+ });
```

The next two follow from the first: both signatures named the engine, so neither
trait could be implemented once it is private.

- `CodeActionProvider` receives an `Entity<EditorState>`. The engine finds it
  through a weak back-reference recorded by `EditorState::new`.

```diff
  fn perform_code_action(
      &self,
-     state: Entity<InputBaseState>,
+     state: Entity<EditorState>,
      action: CodeAction,
      push_to_history: bool,
      window: &mut Window,
      cx: &mut App,
  ) -> Task<Result<()>>;
```

- `InputHighlighter::update` takes a `HighlighterHost` and `&mut App`. A
  background parse needs to write its fold candidates back; `&mut App` cannot do
  that, and `Context<EditorState>` re-enters on the `set_value` path, where the
  editor is already borrowed.

```diff
  fn update(
      &mut self,
      edit: Option<InputEdit>,
      text: &Rope,
      folding: bool,
+     host: HighlighterHost,
      window: &mut Window,
-     cx: &mut Context<InputBaseState>,
+     cx: &mut App,
  );
```

```diff
- let _ = entity.update(cx, |state, cx| {
-     state.apply_highlighter_fold_candidates(folds, cx);
- });
+ let _ = cx.update(|cx| host.apply_fold_candidates(folds, cx));
```

  This also fixes the WASM build: `syntect_highlighter.rs` and the `editor_story`
  WASM branch implement `update` with `&mut App` against a trait that asked for a
  context, and only `cfg(wasm)` compiles them, so the mismatch was never seen.

### Known limit

`InputState::base_state` and friends stay `#[doc(hidden)] pub`: `gpui-base` and
`gpui-component` are separate crates, so `pub(crate)` is not available, and the
UI layer needs the engine entity to render. The type is unnameable from an
application and absent from the docs, but not sealed.

### DIFF
--- a/crates/base/examples/showcase/syntect_highlighter.rs
+++ b/crates/base/examples/showcase/syntect_highlighter.rs
@@ -1,8 +1,8 @@
 use std::{collections::HashMap, ops::Range, sync::LazyLock};
 
-use gpui::{Context, HighlightStyle, SharedString, Window, rgb};
+use gpui::{App, HighlightStyle, SharedString, Window, rgb};
 use gpui_base::input::{
-    FoldRange, HighlightStyleResolver, InputBaseState, InputEdit, InputHighlighter, Rope,
+    FoldRange, HighlightStyleResolver, HighlighterHost, InputEdit, InputHighlighter, Rope,
 };
 use syntect::{
     parsing::{ParseState, Scope, ScopeStack, SyntaxReference, SyntaxSet},
@@ -71,8 +71,9 @@ impl InputHighlighter for SyntectHighlighter {
         _edit: Option<InputEdit>,
         text: &Rope,
         folding: bool,
+        _host: HighlighterHost,
         _window: &mut Window,
-        _cx: &mut Context<InputBaseState>,
+        _cx: &mut App,
     ) {
         // `syntect` has no incremental mode, so the whole document is reparsed.
         // Read the rope once and reuse that string for folding too.
--- a/crates/base/src/input/base/mode.rs
+++ b/crates/base/src/input/base/mode.rs
@@ -253,7 +253,8 @@ impl InputMode {
                     update.selected_range,
                     update.change_text,
                 );
-                h.update(Some(edit), update.new_text, *folding, window, cx);
+                let host = crate::input::HighlighterHost::new(cx.weak_entity());
+                h.update(Some(edit), update.new_text, *folding, host, window, cx);
             }
             _ => {}
         }
--- a/crates/base/src/input/base/state.rs
+++ b/crates/base/src/input/base/state.rs
@@ -9,7 +9,7 @@ use gpui::{
     EventEmitter, FocusHandle, Focusable, InteractiveElement as _, IntoElement, KeyBinding,
     KeyDownEvent, MouseButton, MouseDownEvent, MouseMoveEvent, MouseUpEvent, ParentElement as _,
     Pixels, Point, Render, ScrollHandle, ScrollWheelEvent, SharedString, Styled as _, Subscription,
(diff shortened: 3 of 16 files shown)


huacnlee:
Superseded by #2716, which takes the typestate route instead.

The facade route hit a wall: with `EditorState` as a separate entity, a reader either takes a `cx` (breaking 85+ call sites) or caches (a second copy of the whole text, and it goes stale). #2716 makes the three states aliases of one engine, so `value()` reads `&self` — no `cx`, no copy, nothing to sync — and the LSP seams need neither the weak back-reference nor the `HighlighterHost` this PR introduced.

PR #2752, opened 2026-08-18
text: Add `table_actions` hook for Markdown tables

madcodelife:
## Summary

- Add a `TextView::table_actions` hook, mirroring the existing `code_block_actions`: apps can render their own element (copy / download buttons, …) directly below every Markdown table.
- The hook receives a plain `TableData` snapshot — header cells, body rows, the table re-serialized to GFM Markdown, and the source byte range — so callers never need access to the internal node types. `TableData` is re-exported from `text`, unlike `CodeBlock`.
- `TableData.span.start` stays stable while a table streams in (only `end` grows), making it a safe element id seed for documents holding several tables.
- Both table layouts (scrolling and wrapping) render the actions row, with a small gap so the buttons' hover backgrounds stay clear of the table border. Alignment is left to the caller.
- Fix `BlockNode::to_markdown` for tables along the way: it joined cells straight from `Paragraph::to_markdown`, which trails `"\n\n"`, and emitted no outer pipes — so single-column tables were not valid GFM. Table serialization now lives in `Table::to_markdown`, mirroring `table_selected_source`.
- The Markdown example demonstrates the hook with a row showing the table shape plus copy-as-Markdown / CSV / TSV actions.

## Test plan

- [x] `cargo test -p gpui-component text` — 129 passed (123 before; 5 new unit tests for `to_markdown` / `table_data`, 1 new render test)
- [x] `table_actions_render_below_the_table` draws a real frame and asserts the painted bounds sit below the table, covering **both** the scrolling and wrapping layouts; verified it is not vacuous by temporarily removing the render hook (test fails as expected)
- [x] `cargo build -p gpui-component`, `cargo clippy -p gpui-component --all-targets`, `cargo check --no-default-features` — clean
- [x] `cargo +nightly fmt --all -- --check` — clean
- [x] `cargo run --example markdown -p gpui-component-story` — actions row renders below tables (including tables nested in list items), reports the correct row × column counts, and copy actions work

### DIFF
--- a/crates/story/examples/markdown.rs
+++ b/crates/story/examples/markdown.rs
@@ -19,11 +19,12 @@ use gpui_component::{
         DocumentRangeSemanticTokensProvider, Editor, EditorState, InputEvent, Rope, RopeExt,
         TabSize,
     },
+    menu::{DropdownMenu as _, PopupMenuItem},
     resizable::{h_resizable, resizable_panel},
     status_bar::StatusBar,
     text::{
-        MarkdownNode, MarkdownParseContext, MarkdownPlugin, SelectionFormat, TextViewStyle,
-        markdown, markdown_ast,
+        MarkdownNode, MarkdownParseContext, MarkdownPlugin, SelectionFormat, TableData,
+        TextViewStyle, markdown, markdown_ast,
     },
     v_flex,
 };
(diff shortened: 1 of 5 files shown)


crates/story/examples/markdown.rs
```diff
@@ -1267,6 +1308,78 @@ impl Render for Example {
                                                         }
                                                     })
                                             })
+                                            .table_actions(|table, _window, cx| {
+                                                // The hook hands over the table as plain data:
+                                                // header cells, body rows, and the table
+                                                // re-serialized to GFM Markdown.
+                                                let markdown = table.markdown.clone();
+                                                let csv = table_to_csv(table);
+                                                let tsv = table_to_tsv(table);
```
huacnlee: This line should not be here. 

This method will call on every render.

huacnlee (dismissed):

crates/story/examples/markdown.rs
madcodelife: Done.

PR #2755, opened 2026-08-18
table: Add ranged dumps for batched export

### LINKED ISSUE #2754: TableState::dump eagerly materializes unbounded table data and can exhaust memory
## Agreed Resolution

Following the maintainer discussion in #2755, the compatibility decision is to
keep the existing eager `TableState::dump(cx)` API unchanged, add a public
`dump_range(range, cx)` API for bounded row-range materialization, and update
the DataTable story to export CSV asynchronously in fixed-size batches.

The application layer owns the batching loop and file writer. Progress UI,
cancellation, remote pagination, and delegate redesign remain outside the
initial patch.

The earlier limit-based Suggested Direction below is superseded by this
decision.

## Description

`TableState::dump` trusts the row and column counts returned by `TableDelegate` and eagerly materializes every header and cell into `Vec<String>` / `Vec<Vec<String>>` without a row, cell, or byte limit.

This is inconsistent with normal table rendering, which is virtualized. A large or remotely backed table can therefore be safe to view but unsafe to export. Calling `dump` can freeze the UI, cause severe memory pressure, or abort the process on allocation failure.

The DataTable story exposes the behavior through its **Export CSV** action. The action first calls `table.dump(cx)`, then serializes the materialized rows into an in-memory `csv::Writer<Vec<u8>>`, and only prompts for a destination after the CSV has been built. Consequently, cancelling the file picker does not prevent the expensive work.

`String::from_utf8(data)` reuses the CSV `Vec<u8>` allocation rather than making another full copy. However, the eager table dump and the growing CSV buffer still create substantial overlapping memory usage, and the initial rows allocation alone can abort for a sufficiently large delegate-reported count.

The story's default 5,000-row data set is unlikely to trigger the problem. The concern is the public `TableState::dump` API and downstream applications whose delegates represent large, lazy, remote, or user-influenced data sets.

## Environment

- GPUI: v0.2.2 (`cc053a4a6fa2fd0e8793201ed9099466af1be0b1` at reproduction time)
- GPUI Component: v0.5.2, dynamically reproduced at `81305ef4a0fd86f64777791dd38ead5c303a15f4`
- Current affected `main`: `972a3ebfd01afca7da6d8b6f31c9a51288ea5565` (the table dump path is unchanged)
- Rust: 1.96.1 (`x86_64-unknown-linux-gnu`)
- Platform: Arch Linux, kernel 7.1.7

## Steps to Reproduce

1. Add the following temporary integration test as `crates/ui/tests/table_dump_oom_repro.rs`:

```rust
use gpui::{App, Context, IntoElement, TestAppContext, Window, div};
use gpui_component::table::{Column, TableDelegate, TableState};

struct HugeDelegate;

impl TableDelegate for HugeDelegate {
    fn columns_count(&self, _: &App) -> usize {
        8
    }

    fn rows_count(&self, _: &App) -> usize {
        20_000_000
    }

    fn column(&self, col_ix: usize, _: &App) -> Column {
        Column::new(format!("column-{col_ix}"), format!("Column {col_ix}"))
    }

    fn render_td(
        &mut self,
        _: usize,
        _: usize,
        _: &mut Window,
        _: &mut Context<TableState<Self>>,
    ) -> impl IntoElement {
        div()
    }

    fn cell_text(&self, _: usize, _: usize, _: &App) -> String {
        "x".to_owned()
    }
}

#[gpui::test]
fn unbounded_dump_exhausts_memory(cx: &mut TestAppContext) {
    cx.skip_drawing();
    let window = cx.add_window(|window, cx| TableState::new(HugeDelegate, window, cx));

    window
        .update(cx, |table, _, cx| {
            let _ = table.dump(cx);
        })
        .unwrap();
}
```

2. Compile the test without running it:

```sh
cargo test -p gpui-component --test table_dump_oom_repro --no-run
```

3. On Linux, run the compiled test with a 256 MiB virtual-memory limit:

```sh
test_bin=$(find target/debug/deps -maxdepth 1 -type f -perm -111 \
  -name 'table_dump_oom_repro-*' -print -quit)

(
  ulimit -v 262144
  RUST_BACKTRACE=full "$test_bin" \
    --exact unbounded_dump_exhausts_memory --nocapture
)
```

4. Observe that the process aborts before `cell_text` iteration begins.

The 20,000,000-row value is intentionally used here to make the manual PoC
fail quickly and deterministically under a memory limit. An automated
regression test only needs a small value immediately above `max_rows`; that
keeps a future regression from attempting the full PoC allocation on a CI
runner.

## Expected

`TableState::dump` should require explicit resource budgets and return a
recoverable error before allocating or retaining data beyond those budgets.
The existing CSV action should supply finite limits so its in-memory output can
no longer grow solely from unbounded delegate-reported counts or cell data.

## Actual

The test process aborts with exit status 134 (`SIGABRT`):

```text
running 1 test
memory allocation of 480000000 bytes failed
...
gpui_component::table::state::TableState<D>::dump
    at crates/ui/src/table/state.rs:532:24
```

On a 64-bit target, the outer `Vec<Vec<String>>` requires 24 bytes per row. Therefore `Vec::with_capacity(20_000_000)` immediately requests 480,000,000 bytes before allocating any per-row vectors or cell strings.

Smaller inputs can still consume substantial memory. For example, one million rows with eight columns require roughly 216 MiB for only the outer row vectors and `String` metadata, excluding string contents, per-allocation overhead, headers, and the CSV output buffer.

## Relevant Code

- `crates/ui/src/table/state.rs:521-542`: `TableState::dump` eagerly allocates every row and cell from delegate-reported counts.
- `crates/ui/src/table/delegate.rs:16-26,224-230`: applications control `rows_count`, `columns_count`, and exported cell text.
- `crates/story/src/stories/data_table_story.rs:1108-1157`: CSV is fully generated in memory before prompting for a path.
- `crates/story/src/stories/data_table_story.rs:1474-1478`: the action is exposed through the **Export CSV** button.

## Suggested Direction

Implement explicit limits for the three dimensions that make the current dump
unbounded:

- maximum data rows;
- maximum header and data cells combined, with checked arithmetic;
- maximum cumulative capacity of returned header and cell strings.

Validate row and cell counts before allocating output vectors, use fallible
vector reservations, and return a recoverable error when a limit or allocation
is exceeded. The DataTable story should pass finite limits to its existing CSV
export path.

These three limits address one resource-exhaustion issue rather than separate
adjacent concerns: a row-only limit could be bypassed by a huge column count,
while row and cell limits alone would still allow a small number of very large
strings.

## Non-goals

The smallest complete fix does not need to add or change:

- streaming CSV directly to disk;
- background or concurrent export work;
- cancellation handling;
- file-picker ordering;
- UI error presentation;
- remote pagination or `TableDelegate` architecture;
- disk-space handling.

Those changes may be useful independently, but should be proposed separately
instead of expanding this issue's patch.

## Regression Coverage

An integration test under `crates/ui/tests` should exercise the public API and
cover each introduced budget:

- row count is rejected before any `cell_text` call;
- total cell count prevents a large column count from bypassing the row limit;
- cumulative string capacity is rejected when row and cell counts are small.

Use a small row count immediately above the configured test limit rather than
the 20,000,000-row manual PoC. The large value belongs in the reproduction
above; a small boundary value gives the same deterministic assertion without
risking CI resource exhaustion if the check regresses.

## Screenshots

Not applicable; the failure is an allocation abort. The backtrace and reproduction output above demonstrate the issue.

lurenjia534:
Closes #2754

## Description

`TableState::dump_range` lets callers materialize a selected row range while
preserving the existing full-table `dump` API. The requested range is clamped
to the delegate's current row count so applications can iterate over large
tables in fixed-size batches.

The DataTable story now prompts for the destination before reading table data,
then materializes 2,000 rows at a time on the foreground GPUI context. A
capacity-one channel sends those batches to one background `csv::Writer<File>`,
which streams directly to disk without accumulating the complete table, CSV
buffer, or `String` in memory.

`TableState::dump` remains an eager whole-table API for compatibility. Large
data exporters should use `dump_range` with an application-level batch size.
Progress UI, cancellation, remote pagination, and delegate changes remain
outside this PR.

## Screenshot

Not applicable. This changes export scheduling and file I/O rather than the
rendered table UI.

## How to Test

```sh
cargo fmt --all -- --check
cargo test -p gpui-component --test table_dump_range
cargo test -p gpui-component
cargo check -p gpui-component-story
RUSTC_BOOTSTRAP=1 cargo check \
  -p gpui-component-story-web \
  --target wasm32-unknown-unknown
```

Observed results:

- formatting passed;
- both `table_dump_range` tests passed;
- the complete `gpui-component` test suite passed;
- the native story crate compiled successfully;
- the WASM story check passed after enabling the unstable feature required by
  the current `wasm_thread` dependency (`RUSTC_BOOTSTRAP=1`); a plain stable
  check stops in that dependency with `E0554` before compiling the story;
- the DataTable story opened its save dialog before export work; cancelling
  produced no output, while exporting the default table produced one header
  and 5,000 records with 45 columns each;
- the first exported record had ID `0`, the last had ID `4999`, and the CSV
  parsed successfully with consistent column counts.

The ranged dump regression tests verify that `dump_range(98..103)` is clamped
to a 100-row table and only calls `cell_text` for rows 98 and 99, while the
existing `dump(cx)` still returns the complete table.

## AI Assistance

AI assisted with root-cause analysis, implementation, test design, validation
planning, and PR wording. I reviewed the final diff and ran the checks listed
above.

## Checklist

- [x] Reviewed the changes in this PR and confirmed AI-generated code is accurate.
- [ ] Tested macOS, Windows, and Linux platform performance (not platform-specific).


huacnlee (dismissed):
I'd think add a limit is not a good solution.

In this case, I think we can have a async, batch export to support a large data table. @madcodelife

PR #2836, opened 2026-08-26
input: Emit replayable text-change deltas

Travis-Gilbert:
## Description

Input consumers currently receive `InputEvent::Change`, which reports that the
value changed but not which accepted edit produced it. Integrations that keep
their own document model must diff the full buffer after every event, and that
becomes especially fragile around UTF-16 input coordinates, astral Unicode,
normalization, and masks.

This adds an independent `InputChangeEvent` subscription channel. Each event
contains a UTF-8 byte range into the value before the edit and the replacement
text. Applying those two values reconstructs the value after the edit. When a
mask rewrites surrounding text, the event reports a whole-value replacement so
the same replay invariant remains true.

The existing `InputEvent::Change` event is unchanged and still emitted once per
accepted edit. `set_value` remains silent on both channels. This makes the API
additive for current consumers while allowing document-backed editors to adopt
precise deltas incrementally.

## How to Test

- `CARGO_TARGET_DIR=/Volumes/... cargo test -p gpui-base emits_` (5 passed,
  including the two new delta-event tests)
- `CARGO_TARGET_DIR=/Volumes/... cargo clippy -p gpui-base --lib -- -D warnings`
- `rustfmt --edition 2024 --check crates/base/src/input/base/state.rs`

The new tests cover an astral-Unicode replacement received in UTF-16
coordinates, continued delivery of the legacy change event, silent
`set_value`, and a masked edit that requires a replayable whole-value delta.

## Checklist

- [x] I have read the CONTRIBUTING document and followed the guidelines.
- [x] A human has reviewed and confirmed the AI-assisted code is accurate.
- [ ] Passed `cargo run` for story tests (not run; this change has no visual presentation).
- [x] Platform-specific performance testing is not applicable to this event-only change.

## AI Assistance

OpenAI Codex generated the implementation and tests. The targeted automated
checks above pass; separate human review remains required.

### DIFF
--- a/crates/base/src/input/base/state.rs
+++ b/crates/base/src/input/base/state.rs
@@ -117,6 +117,37 @@ pub enum InputEvent {
     Blur,
 }
 
+/// Describes one accepted change to an input's text.
+///
+/// The range contains UTF-8 byte offsets into the value immediately before the
+/// change. Replacing that range with [`Self::replacement`] produces the value
+/// immediately after the change. Inputs whose masks rewrite surrounding text
+/// report a whole-value replacement so the same invariant still holds.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct InputChangeEvent {
+    range: Range<usize>,
+    replacement: SharedString,
+}
+
+impl InputChangeEvent {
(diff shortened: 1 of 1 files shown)


huacnlee (changes requested):
Please read my last reply to change.

PR #2865, opened 2026-08-28
table: a mouse click on a cell no longer scrolls it

emakarov:
## Summary

`Table::on_cell_click` selects the clicked cell via `set_selected_cell`, which unconditionally calls `vertical_scroll_handle.scroll_to_item(row_ix, ScrollStrategy::Center)`. A mouse click can only land on a cell the user can already see, so forcing it to the centre of the viewport on every click recentres the table under the pointer for no reason. This is most noticeable on a short/medium table, where the clicked row visibly jumps.

Keyboard and programmatic selection are unaffected: `set_selected_cell`'s default behaviour (`scroll: true`) is unchanged, since either can legitimately land on a cell that's off screen and does need scrolling into view.

## Change

- Extracted `set_selected_cell`'s body into `set_selected_cell_with_scroll(row_ix, col_ix, scroll, cx)`, with `set_selected_cell` calling it with `scroll: true` (unchanged public behaviour).
- `on_cell_click` now calls `set_selected_cell_with_scroll(row_ix, col_ix, false, cx)`.

## Test plan

- [x] `cargo check -p gpui-component` passes
- [x] Verified in a downstream app: clicking rows in a table (including the very first, already-visible row) no longer scrolls the viewport; keyboard navigation (arrow keys, Home/End, PageUp/PageDown) still scrolls correctly when moving to an off-screen row.

### DIFF
--- a/crates/ui/src/table/state.rs
+++ b/crates/ui/src/table/state.rs
@@ -486,13 +486,33 @@ where
     /// });
     /// ```
     pub fn set_selected_cell(&mut self, row_ix: usize, col_ix: usize, cx: &mut Context<Self>) {
+        self.set_selected_cell_with_scroll(row_ix, col_ix, true, cx);
+    }
+
+    /// [`Self::set_selected_cell`], with the scroll made optional.
+    ///
+    /// A mouse click can only land on a cell the user can already see, so
+    /// forcing it to the centre of the viewport moves content out from under
+    /// the pointer for no reason — `on_cell_click` selects with `scroll:
+    /// false` for exactly that case. Keyboard and programmatic selection
+    /// keep `scroll: true`, since either can land on a cell that is off
+    /// screen.
+    fn set_selected_cell_with_scroll(
+        &mut self,
+        row_ix: usize,
+        col_ix: usize,
+        scroll: bool,
+        cx: &mut Context<Self>,
+    ) {
         self.selection_mode = SelectionMode::Cell;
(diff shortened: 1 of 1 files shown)


huacnlee (changes requested):
Thanks for digging into this. The jump is real, but the cause is different from the description, and that changes the fix.

`Center` is not applied on every click. In gpui's `uniform_list` the scroll block sits behind a visibility check:

```rust
if scroll_strict || is_above || is_below {
    match strategy { Top => …, Center => …, Bottom => …, Nearest => {} }
}
```

and `scroll_to_item` passes `scroll_strict: false`. So on a fully visible row, `Center` already does nothing today. What jumps is a row **partly** cut off at an edge: `is_below` is true, and `Center` pulls it to the middle.

Not scrolling is not right for that case either — you select a row you still cannot see in full. gpui has what you want as `ScrollStrategy::Nearest`: nothing when the row is visible, minimal reveal when it is not. So this can be one line:

```diff
-        .scroll_to_item(row_ix, ScrollStrategy::Center);
+        .scroll_to_item(row_ix, ScrollStrategy::Nearest);
```

which also fixes the keyboard — arrow and Page keys go through `set_selected_cell` too, so stepping one row past the bottom edge recenters the whole table. This PR keeps that.

One question: you mention the first, already-visible row scrolling. Per the code above that should only happen if the table was scrolled and the row was clipped. If it reproduces on a fully visible row, something else is going on and I would rather find it than cover it up.

PR #2905, opened 2026-09-01
a11y: Expose sidebar items and keep input semantics on the focus node

Pates2004:
`SidebarMenuItem` and `Input` currently render the visible control separately
from the node that owns keyboard focus. On Windows this leaves NVDA/UI Automation
with unnamed, non-focusable settings entries or causes AccessKit to fall back to
announcing the whole window.

## Changes

- render sidebar menu entries as semantic `gpui_base::Button`s with stable names,
  keyboard focus, and Enter/Space activation;
- expose a named caret button for expandable entries and keep caret clicks from
  activating the parent item;
- project input role, label, placeholder, value, and SetValue onto the focused
  editor node;
- keep the decorative input frame as a non-interactive group so the tree has a
  single actionable text field.

## Testing

- `RUSTFLAGS=-D warnings cargo test -p gpui-component input::input::tests --lib`
- `RUSTFLAGS=-D warnings cargo test -p gpui-component sidebar --lib`
- Windows UI Automation smoke test with NVDA running: Settings entries are
  named buttons and the search field is a named Edit control.

Real hardware was not required for this library-level change; OpenLogi hardware
verification remains pending.

### DIFF
--- a/crates/base/src/input/base/state.rs
+++ b/crates/base/src/input/base/state.rs
@@ -4,11 +4,12 @@
 //! https://github.com/zed-industries/zed/blob/main/crates/gpui/examples/input.rs
 use gpui::TextAlign;
 use gpui::{
-    Action, App, AppContext, Bounds, ClipboardItem, Context, Edges, Entity, EntityInputHandler,
-    EventEmitter, FocusHandle, Focusable, InteractiveElement as _, IntoElement, KeyBinding,
-    KeyDownEvent, MouseButton, MouseDownEvent, MouseMoveEvent, MouseUpEvent, ParentElement as _,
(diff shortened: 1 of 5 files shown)


huacnlee (changes requested):
This PR currently mixes two independent changes: Sidebar accessibility and Input accessibility. Our contribution guidelines strictly require one PR to do one thing. Please keep this PR focused on the Sidebar accessibility fix and move the Input changes to a separate PR.

The Sidebar refactor also introduces a visible layout regression: submenu labels and controls overlap, as shown in the screenshot. Please restore the existing layout and add regression coverage for expanded and collapsed Sidebar states.

image

The unrelated Polish locale additions should also be removed or submitted separately.

PR #2948, opened 2026-09-04
text_selection: drag-autoscroll never moved a virtualized list

kossoy:
A drag held at the bottom of a padded, scrollable `TextView` can leave its virtualized list stationary. The synthetic wheel position is clamped inside the nearest content mask, but the inner List ends above that edge. The event then lands in the padding, outside the List's hitbox.

This change sends auto-scroll to self-scrolling participants through their own list API and reserves synthetic wheel events for participants that scroll through an ancestor. Dispatch is exclusive, so exactly one path drives the list.

## Reproduction and measurements

Use a clipped row containing a `scrollable(true)` TextView with `py_3()`. Anchor a selection in the visible text, drag to the clipping edge and hold.

In the padded harness on base `928c3eb776a3d733d9b771f7dea27a6a79242ced`, the List ends at y=288 while the content mask ends at y=300. Over four ticks, a pointer at y=286, inside the List, moves it 52.33 px; y=298, in the padding, moves it 0 px.

The [maintainer's measurement](https://github.com/longbridge/gpui-kit/pull/2948#issuecomment-5540782842) and [reproduction and rework results](https://github.com/longbridge/gpui-kit/pull/2948#issuecomment-5551008307) establish the cause and check the scroll magnitude:

| Revision | Movement after 64 ms / four ticks |
|---|---:|
| Base, wheel only, pointer inside List | 78.33 px |
| Initial PR, participant plus wheel | 156.67 px |
| Current head, exclusive participant dispatch | 78.33 px |

Synthetic wheel events work when their position hits the List. Pointer capture is not the cause of the padded-reader failure.

## Implementation

- `TextSelectionRegistration` carries a `self_scroll` flag, set from `TextView::scrollable`.
- `update_auto_scroll` routes self-scrolling participants through `update_participant_auto_scroll` and returns. That path measures against the participant's own bounds; only the ancestor-wheel path uses the content mask.
- `update_auto_scroll` takes `&Window`; `anchor_participant` and `anchor_registration` share the anchor lookup.
- The existing both-directions harness asserts a magnitude bound. `padded_reader_text_view_auto_scrolls_at_the_clipping_edge` covers the padding failure.

Changes are in base text selection and the component selection harness. GPUI core is unchanged.

## Validation

For current head `bee77eaa61ae23e5661a07edb9913e708f3b223c`, the [rework results](https://github.com/longbridge/gpui-kit/pull/2948#issuecomment-5551008307) record:

- `cargo check -p gpui-base -p gpui-component`.
- `cargo test -p gpui-component --lib window_selection`: 52 passed.
- Base `text_selection`, `text::`, and `auto_scroll` tests: 150 passed.
- Formatting and Clippy checks reported clean for the touched files.

All nine GitHub check runs on this head reported success when inspected on 2026-09-08. Maintainer review of the reworked head remains pending.

### DIFF
--- a/crates/base/src/text/selection_adapter.rs
+++ b/crates/base/src/text/selection_adapter.rs
@@ -190,12 +190,14 @@ impl TextViewSelectionAdapter {
         self.text_bounds.extend(bounds);
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub(super) fn register(
         &self,
         hitbox: Hitbox,
         bounds: Bounds<Pixels>,
         scroll_offset: Point<Pixels>,
         document_order: u64,
+        self_scroll: bool,
         window: &mut Window,
         cx: &mut App,
     ) {
@@ -203,7 +205,8 @@ impl TextViewSelectionAdapter {
             TextSelectionRegistration::new(hitbox, bounds)
                 .with_scroll_offset(scroll_offset)
                 .with_document_order(document_order)
-                .with_text_bounds(self.text_bounds.clone()),
+                .with_text_bounds(self.text_bounds.clone())
+                .with_self_scroll(self_scroll),
             window,
             cx,
         );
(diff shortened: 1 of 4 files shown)


huacnlee (changes requested):
check again

huacnlee (approved):

PR #3039, opened 2026-09-11
highlighter: allow languages to register runtime parser factories

feigeCode:
## Problem

A language's Tree-sitter grammar has to be statically linked into the binary before
highlighting can use it. There's no way to plug in a grammar that arrives at runtime —
WASM-parsers shipped and updated separately, for example. `LanguageRegistry` stores a map
of `GrammarConfig`, so everything downstream assumes a compiled-in `Language`.

## What this does

Adds an optional per-language parser factory that takes precedence over the static grammar:

- `LanguageParserFactory` — `Arc<dyn Fn() -> Result<(Parser, Language)> + Send + Sync>`.
- `LanguageRegistry::register_parser_factory(lang, factory)` — registers one for a language.
- `LanguageRegistry::parser(name)` — resolves a parser + grammar, preferring a registered
  factory and falling back to the statically linked grammar.

The three places that previously built `Parser` themselves now go through `parser(...)`:

- `SyntaxHighlighter::new`
- the "parse only this range" path used for partial re-highlighting
- `TreeSitterInputHighlighter` in `input_adapter.rs`

Nothing changes for callers that don't register a factory: no factory ⇒ the old static
grammar path, exactly as before.

## Why the factory returns both parser and grammar

A dynamic grammar often wants its own `Parser` instance (its own wasm store / cancellation
token). Returning `(Parser, Language)` as a pair lets the factory decide how the parser is
configured instead of forcing every kind of grammar into one pre-built `Parser`.

## Validation

- `cargo check -p gpui-component --no-default-features` (tree-sitter off) passes, so the
  `#![cfg(feature = ...)]` boundaries around the highlighter are unchanged.
- `cargo clippy -p gpui-component -p gpui-component-story -p gpui-kit-assets -- --deny warnings`
  passes.
- `cargo test --workspace --exclude gpui-shell --features gpui-component-story/test-support`
  passes — the existing highlighter tests exercise all three call sites.

No new unit test is included: asserting that a factory wins would require constructing a
tree-sitter `Language`, which can't be done without a real grammar fixture. If you'd rather
have coverage here, tell me how you'd like it faked and I'll add it.

Downstream, we use this to drive highlighting from grammars loaded at runtime.

Release Notes:

- N/A

### DIFF
--- a/crates/component/src/highlighter/highlighter.rs
+++ b/crates/component/src/highlighter/highlighter.rs
@@ -380,12 +380,14 @@ impl SyntaxHighlighter {
 
         // Languages without grammar default to a highlighter that never
         // parses and creates no styles.
-        let Some(grammar) = config.language.as_ref() else {
+        let Some(_) = config.language.as_ref() else {
             return Ok(Self::build_inert(config.name.clone()));
         };
 
-        let mut parser = Parser::new();
-        parser.set_language(grammar).context("parse set_language")?;
+        let (mut parser, grammar) = LanguageRegistry::singleton().parser(lang)?;
+        parser
+            .set_language(&grammar)
+            .context("parse set_language")?;
 
         // Concatenate the query strings, keeping track of the start offset of each section.
         let mut query_source = String::new();
@@ -397,7 +399,7 @@ impl SyntaxHighlighter {
 
         // Construct a single query by concatenating the three query strings, but record the
         // range of pattern indices that belong to each individual string.
-        let mut query = Query::new(grammar, &query_source).context("new query")?;
+        let mut query = Query::new(&grammar, &query_source).context("new query")?;
 
         let mut locals_pattern_index = 0;
         let mut highlights_pattern_index = 0;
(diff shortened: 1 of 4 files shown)


huacnlee (changes requested):
Two blocking issues, detailed in https://github.com/longbridge/gpui-kit/pull/3039#issuecomment-5631679339:

1. `LanguageRegistry::parser()` invokes the factory while the `parser_factories` mutex guard is still alive (the guard is an `if let` scrutinee temporary). Re-entrant factories deadlock, every `parser()` call is serialized behind the factory, and a panicking factory poisons the singleton. Bind the `.cloned()` result in its own `let` before the `if let`.

2. `build_for_language` still gates on the static `config.language` before calling `parser()`, so a factory-only language (the runtime-loaded case this PR is for) gets an inert highlighter and the factory is never called. `resolve_language` and `input_adapter.rs`'s `has_grammar()` carry the same static-only assumption. Please add a test that registers a language with `language: None` plus a factory and goes through `SyntaxHighlighter::new`.

huacnlee (approved):

PR #3053, opened 2026-09-13
text: let a TextViewStyle choose the inline-code font scale

0xR32:
### The problem

Inline code is drawn at `0.875` of the surrounding text, hardcoded in `mark_highlight` (`crates/base/src/text/node.rs`). A GPUI run may vary its font but not its size, so `text_size_ranges` must split any paragraph containing a code span into separately shaped lines.

That works while every span fits on one line. A span long enough to **wrap inside its own piece** is laid out against a line the other pieces never agreed to:

- its background quad is sized by one piece and its continuation positioned by another, so the span's background runs past its glyphs and the tail lands at the wrong x;
- the paragraph paints taller than it measured, so it overlaps the block below it.

Short spans in the same paragraph render perfectly, which is why this looks intermittent — it tracks span length against the available width.

### Why not just remove the scale

The shrink is a good default: a monospace span really does read larger than proportional prose beside it. So this changes nothing by default.

It is the wrong default for an application whose body text is *already* monospace. There the span is the same face as its neighbours, the scale buys nothing visually, and all it does is force the split that breaks wrapping.

### The change

`TextViewStyle`'s own docs say the fields are private so that "a later field is an additive change rather than a breaking one", so this follows that pattern:

- `TextViewStyle::with_inline_code_font_scale(Option<f32>)` and a matching accessor
- a public `INLINE_CODE_FONT_SCALE` constant (`0.875`) as the default
- `mark_highlight` reads the style instead of the literal

Passing `None` draws inline code at the surrounding size, which keeps the paragraph a single shaped line and lays out correctly at every width.

**No existing behaviour changes** — any caller that does not set it renders exactly as before.

### Tests

`cargo test -p gpui-base --lib text::` — 143 passed, including a new `inline_code_is_shrunk_by_default_and_can_be_left_alone` asserting both the preserved default and the override.

### Note

This is a knob over the symptom, not a fix for the underlying layout bug — a code span that wraps inside its own shaped piece is still mislaid out when a scale *is* set. Happy to look at that separately if you'd rather have it fixed at the source; this was the smallest non-breaking change that unblocks monospace-bodied apps.


huacnlee (changes requested):
`with_inline_code_font_scale` is not a good API design. It's most like a hard code.

PR #3083, opened 2026-09-15
virtual_list: Keep a frame's cost independent of the item count

huacnlee:
Every frame `VirtualList` copied its per-item sizes and origins out of the element state, allocated a `Bounds` for every item to serve a deferred scroll-to-item, and walked the items from the front to find the visible range. At 10k items that is under one percent of a frame; at 1M items it is most of it.

- `ItemSizeLayout` keeps `sizes` and `origins` behind an `Rc<[Pixels]>`, so carrying them between the element state and the frame state is a reference count. `content_size` is the running total from the same pass instead of a second sum.
- The visible range is `visible_range(origins, sizes, viewport)`: two binary searches over the item ends, which grow with the index. Same result as the linear scans (leading padding admitted at the start, one item of overdraw past the end), with tests for the edges: first/last item, uneven sizes, empty list, viewport past the content.
- A deferred scroll-to-item computes the bounds of the one item it targets (`ItemSizeLayout::item_bounds`) instead of every item's.

Scrolling a full-screen list of 24 visible rows at 144 Hz (Linux, release, process CPU on the single-core scale):

| items | `main` | this PR |
|---|---|---|
| 10k | 14.2% | 14.0% |
| 100k | 15.2% | 13.8% |
| 1M | 34.6% | 13.8% |

`cargo test -p gpui-base --lib`: 506 passed (4 new). `cargo clippy -p gpui-base --all-targets -- --deny warnings` clean.


huacnlee:
Folded into #3085.

PR #3084, opened 2026-09-15
theme: Name the family the system font resolves to

huacnlee:
The UI family defaults to the virtual `.SystemUIFont`, which GPUI maps to a platform name (`IBM Plex Sans` on Linux) and, when that is not installed, resolves through its own fallback stack. The fallback works, but every text run of every frame pays for it: GPUI caches the failed lookup as an error and formats and allocates it again each time it is hit (`TextSystem::font_id` → `anyhow!("{err}")`), once per missing entry of the stack, before the installed family answers. On a full-screen list of 24 text rows scrolling at 144 Hz that is 3% of the main thread in `font_id`, `anyhow` and `fmt::write`.

- `theme/system_font.rs`: `Theme::change` asks GPUI once which family `.SystemUIFont` lands on (`resolve_font` + `get_font_for_id`) and, when that is an installed family rather than the system font itself, names it on `Theme::font_family`, so every later lookup hits the cache. No family names are hardcoded: the substitute is whatever GPUI's own fallback picked. A family the application or a theme file set is used as-is, like the monospace probe.
- `mono_font::installed_font_names` reads the installed font list once per process and is shared by both probes (it was already cached by the monospace one; enumerating costs ~100ms on macOS).
- Docs: `fonts.md` (en + zh-CN) describes both probes.

Linux (Adwaita Sans is what `.SystemUIFont` falls back to here): the frames are pixel-identical — `magick compare` on screenshots from `main` and this branch reports 0 differing pixels — and the list scrolls at 12.9% CPU instead of 14.2%. macOS and Windows resolve the system font to itself, so nothing changes there; the test text system lists no fonts, so tests are unaffected either.

`cargo test -p gpui-component --lib`: 933 passed (3 new). `cargo clippy -p gpui-component --all-targets -- --deny warnings` clean.


huacnlee:
Folded into #3085.

PR #3104, opened 2026-09-16
setting: Preserve selected page identity when searching

### LINKED ISSUE #3089: Settings search loses selected page identity after filtering
## Description

Filtering `Settings` rebuilds the page/group arrays, but the stored selection still refers to the previous indexes. As a result, a search can leave the content area blank even when the sidebar contains a matching page, or change which page an existing index refers to.

This report concerns selection consistency during filtering, not additional search keywords or an external page-change callback.

## Environment

- Native reproduction: the full `gpui-component-story` gallery from `main` at [`f29e0490`](https://github.com/longbridge/gpui-kit/commit/f29e049009a7d9cfd6c9e99bcf00948a2c32432f), built with `--no-default-features` and packaged as a macOS `.app`.
- GPUI Component: `0.6.1`; GPUI dependency: `gpui-pre 0.3.5`.
- Platform: macOS 27.0.
- The screenshots below were captured through Computer Use from the unmodified Settings implementation at that revision. The temporary app only adjusts startup to use the checkout as its working directory and open the Settings story by default.
- Earlier source inspection also found the same filtering logic in published `0.6.0` and `0.6.1`. No platform-specific cause is claimed.

## Steps to Reproduce

In the full Story gallery, open the **Settings** story and use the search field inside the Settings panel:

1. With the Settings search empty, select **Software Update**. The page displays **Enable Notifications** and **Auto Update**.
2. Enter `notifications` in the Settings search field.
3. **Software Update** remains as the only matching sidebar page, but the content area becomes blank and the sidebar page loses its selected highlight.
4. Click the remaining **Software Update** result. Its content reappears, showing the matching **Enable Notifications** item.
5. Clear the Settings search. The selected page unexpectedly changes to **General**, even though **Software Update** is still present.

Steps 1–5 were reproduced in the native full Story app. All four screenshots below are from the same unfixed baseline.

## Screenshots

**1. Before searching: Software Update is selected and its content is visible.**

Software Update before searching

**2. Searching for notifications: Software Update matches, but the content area is blank.**

Matching sidebar page with blank content

**3. Clicking the remaining result restores the matching content.**

Clicking Software Update restores Enable Notifications

**4. Clearing the search incorrectly switches the selected page to General.**

Clearing the search jumps to General

## Minimal builder example

Use two pages whose items have distinct keywords, for example this Settings builder in an initialized component application:

```rust
use gpui_kit::{div, ParentElement};
use gpui_kit::component::setting::{
    SettingGroup, SettingItem, SettingPage, Settings,
};

Settings::new("settings-search")
    .page(
        SettingPage::new("General").group(
            SettingGroup::new().item(
                SettingItem::render(|_, _, _| div().child("Language"))
                    .keywords(["language"]),
            ),
        ),
    )
    .page(
        SettingPage::new("Appearance").group(
            SettingGroup::new().item(
                SettingItem::render(|_, _, _| div().child("Theme"))
                    .keywords(["theme"]),
            ),
        ),
    )
```

1. With an empty search, select **Appearance**, the second page.
2. Enter `theme` in the settings search input.
3. Only Appearance matches. Its original index is `1`, but its filtered index is now `0`.
4. The stored selection remains `1`, so no filtered page is rendered.

The exact builder snippet above is a source-derived minimal recipe and was not run separately; native verification used the existing full Story scenario described above.

## Expected

- Keep the selected page when it still matches the search, regardless of its position in the filtered list.
- When the selected page no longer matches, select a matching page consistently, such as the first result.
- Keep sidebar selection, rendered content, and group navigation aligned as the query changes or is cleared.

## Actual

The source follows these steps:

1. [`filtered_pages`](https://github.com/longbridge/gpui-kit/blob/f29e049009a7d9cfd6c9e99bcf00948a2c32432f/crates/component/src/setting/settings.rs#L112-L143) clones and removes pages, groups, and items without retaining their original indexes.
2. [`render`](https://github.com/longbridge/gpui-kit/blob/f29e049009a7d9cfd6c9e99bcf00948a2c32432f/crates/component/src/setting/settings.rs#L379-L387) computes the filtered pages without reconciling `selected_index`.
3. [`render_active_page`](https://github.com/longbridge/gpui-kit/blob/f29e049009a7d9cfd6c9e99bcf00948a2c32432f/crates/component/src/setting/settings.rs#L145-L164) compares the stored page index against an enumeration of that shorter list and returns an empty `div` if none matches.

In the two-page case, this directly produces the blank-content condition despite a matching result. With more pages, a still-valid numeric index can instead refer to a different page after filtering. The sidebar also uses filtered page/group indexes, so fixing only the blank fallback would not address selection identity throughout navigation.

Prepared with AI assistance. The full Story behavior above was verified through native UI interaction; the source links explain the page/group index mismatch.

## Proposed fix: native verification

Verified the local fix on branch `codex/3089-settings-search-selection`, based on `f29e0490`, in the full Story gallery packaged as a macOS `.app` with the same `--no-default-features` build configuration. This is verification of a local working-tree change, not a released fix.

The fix retains original page/group/item indexes while filtering visibility. It preserves the current page if it still matches; otherwise it selects the first matching page. Group navigation translates the original group index to its visible scroll position.

Repeating the native reproduction above now gives:

- Searching for `notifications` keeps **Software Update** selected and immediately displays **Enable Notifications**.
- Clicking that result and then clearing the search keeps **Software Update** selected, with both settings visible.
- Searching for `font` from Software Update selects the matching **General** page.
- An unmatched query shows no page content; clearing it restores the retained General selection.

The three focused regression tests also pass, covering page/group/item identity, sidebar navigation and group scrolling, fallback/no-result recovery, and resetting only matching settings.

**After: searching keeps the matching current page and its content.**

After fix: notifications keeps Software Update selected and visible

**After: clicking the result and clearing search keeps Software Update selected.**

After fix: clearing search retains Software Update

suxiaoshao:
Closes #3089

## Description

Searching Settings currently renumbers the filtered page/group arrays while selection retains its previous numeric index. Selecting **Software Update** and searching for `notifications` therefore leaves the content blank; clicking that result and clearing the search then selects **General**.

Keep page, group, and item indexes in the original data and filter visibility separately. Preserve the current page while it matches, otherwise select the first matching page. Preserve a matching selected group, translate its original index to the visible position for scrolling, and invalidate list measurements when the query or visible groups change. An empty result retains selection for when results return. Reset All continues to affect only matching settings.

Adds three focused regression tests and documents the selection behavior in English and Chinese. Public API signatures are unchanged.

## Screenshot

Captured through Computer Use in the full native Story gallery on macOS 27.0. Before uses `main` at `f29e0490`; after uses this branch's Settings changes. Both were built with `--no-default-features` and packaged as `.app` bundles; temporary startup adjustments only set the working directory and open the Settings story.

| Scenario | Before | After |
| --- | --- | --- |
| Select Software Update, then search `notifications` | Before: matching page but blank content | After: matching page stays selected and visible |
| Click the result, then clear the search | Before: selection jumps to General | After: selection stays on Software Update |

## How to Test

Passed:

- `cargo test -p gpui-component --lib setting::settings::tests --locked` — 3 tests covering page/group/item identity, sidebar clicks, group scrolling, fallback/no-result recovery, and reset scope.
- `cargo clippy -p gpui-component --lib --tests --locked -- -D warnings`
- `cargo fmt --all -- --check`
- Full Story build with `cargo build -p gpui-component-story --no-default-features --locked`.
- Native interaction in the packaged full Story app: select Software Update, search `notifications`, click the result, then clear. Selection and content remain aligned. Searching `font` falls back to General; entering an unmatched query and clearing restores the retained page.

To reproduce, run `cargo run -p gpui-component-story --no-default-features -- settings` and repeat the interactions above. Native verification was performed on macOS; Windows and Linux were not run locally.

## Checklist

- [x] Read the contribution guidelines and kept the change scoped to Settings search selection.
- [x] Reviewed the implementation and ran the focused checks listed above.
- [x] Tested the affected story in the full native Story gallery.
- [ ] Human review of AI-generated changes.

AI assistance: the implementation, regression tests, and documentation changes were generated and reviewed with Codex; native interactions and screenshots were captured through Computer Use.


huacnlee:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #3129, opened 2026-09-18
text: make the gap below a heading configurable

wweziza:
## What

Add `TextViewStyle::heading_gap` (getter) and `with_heading_gap` (setter), mirroring the existing `paragraph_gap`, and use it for the heading's bottom padding in the block renderer.

## Why

Headings render with a hard-coded `.pb(rems(0.3))`, so stacked headings, and a heading above body text, always sit tight no matter what spacing the surrounding view wants. There is no way to loosen it from `TextViewStyle` the way `paragraph_gap` already allows for paragraphs.

## Behavior

The default is `rems(0.3)`, identical to the current constant, so existing output is unchanged. Callers that want roomier headings can now do:

```rust
TextView::markdown(id, text, window, cx)
    .style(TextViewStyle::default().with_heading_gap(rems(1.)))
```

## Notes

One focused change following the `paragraph_gap` pattern:
- `crates/base/src/text/style.rs`: field, PartialEq arm, default, `with_heading_gap`, `heading_gap()`.
- `crates/base/src/text/node.rs`: heading uses `node_cx.style.heading_gap()` instead of the constant.


huacnlee (changes requested):
The proposed `heading_gap` API is too ambiguous for a public styling seam. From the name alone, callers cannot tell whether it controls spacing before a heading, after it, between adjacent headings, or between a heading and body text; the implementation happens to map it to bottom padding, but that directional detail is hidden in `node.rs`.

Please avoid adding another top-level scalar to the already mixed `TextViewStyle`. Instead, consolidate the current heading-specific customization behind a level-aware heading refinement API, for example a `with_heading_style` callback that receives the heading level and returns a `StyleRefinement`. Treat the returned value as an overlay on the default H1-H6 refinement so callers can change spacing, font size, weight, color, or other supported properties without restating defaults. The existing default size, weight, and 0.3rem bottom spacing should move into that default heading refinement rather than remain hard-coded in the renderer. Please also define the compatibility path for the existing heading font-size APIs rather than creating two overlapping ways to style headings.

Add focused coverage proving that per-level refinements affect rendered heading geometry/style while unchanged levels retain their defaults, and document the final Base public API in the PR description. Security: PASS for the affected styling boundary.

Please address the findings above, then request my review again.

